PR #102, opened 2014-10-14
Change names of files with the same path - this time by Transformer

amorfis:
Hello,

What about this one? ;)


ugoscaiella:
Hi,
What's the status of this PR?

I was trying to restore the possibility to filter-out files from certain jars, that was supported in past (according to the old readme, but I never tested it):
```
shadow {
    filter('org.apache.maven.its.shade.fac:a') {
        include '**/a.properties'
    }
    filter('org.apache.maven.its.shade.fac:b:client') {
        exclude 'org/apache/*'
        exclude 'org/apache/maven/b/'
    }
    filter('*:*') {
        exclude 'org/*'
    }
}
```

I tried to implement this using a Transformer, but I think that this PR is required, because the transformer API do not have any information about the path or the name of the jar which the file comes from...
is it right?

Goooler:
Looks this should be handled by

- #546
- #1233
- #1548
- Something missing?

PR #126, opened 2015-03-16
Displaying `warn` level messages for file collisions

chapmajs:
We experienced problems in a recent project where Java property files were being left out due to collisions with other JARs in the project. In the course of debugging the issue, we tried the [OneJar Gradle plugin](https://github.com/rholder/gradle-one-jar) which printed debug messages about these collisions. This PR adds the same functionality to the Shadow plugin.

With warnings about file collisions, it's trivial to add appropriate appending transformers to the Gradle task for the shadow JAR. One interesting thing we noticed with this was frequent collisions on `META-INF/LICENSE`, which has possibly interesting ramifications for some projects, e.g. anything that includes an Affero GPL license.

It was necessary to modify the `visitedFiles` data structure to allow storing the source JAR and file size without ripping through JAR files again. We provided several overloaded methods to allow insertion into `visitedFiles`.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
@@ -132,7 +132,7 @@ public class ShadowCopyAction implements CopyAction {
         private final PatternSet patternSet
         private final ShadowStats stats
 
-        private Set<String> visitedFiles = new HashSet<String>()
+        private Map<String, Map> visitedFiles = new HashMap<String, Map>()
 
         public StreamAction(ZipOutputStream zipOutStr, List<Transformer> transformers, List<Relocator> relocators,
                             PatternSet patternSet, ShadowStats stats) {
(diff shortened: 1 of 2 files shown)


chapmajs:
It appears that #102 attempts to overcome the symptom automatically, this PR would probably help end users identify those problems.

johnrengelman:
Thanks for the contribution. I'll try and take a look soon on this (I'm pretty wrapped up in work at the moment though. It'll probably be awhile).

Could you take a look at adding a test for this? The test fixtures should give you access to the entire console output of the build and you should be able to parse it for the warning message.

chapmajs:
Tests added! Thanks for taking time to handle this PR.

chapmajs:
Is there anything preventing this PR from being merged in?

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
```diff
@@ -157,8 +157,25 @@ public class ShadowCopyAction implements CopyAction {
                     fileDetails.relativePath.pathString.endsWith('.zip')
         }
 
-        private boolean recordVisit(RelativePath path) {
-            return visitedFiles.add(path.pathString)
+        private boolean recordVisit(path, size, originJar) {
```
johnrengelman: please using typing on arguments.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
```diff
@@ -157,8 +157,25 @@ public class ShadowCopyAction implements CopyAction {
...
+
+            if (originJar == null) {
+                originJar = ""
+            }
+
+            visitedFiles.put(path.toString(), [size: size, originJar: originJar])
+            return true
+        }
+
+        private boolean recordVisit(path) {
```
johnrengelman: Argument typing.

PR #214, opened 2016-05-09
Fix table markdown in README

chrisalice:
Fix the markdown in Latest Test Compatibility so GitHub generates a table.


johnrengelman:
Thanks. Pushed that and didn't check it.

Issue #229, opened 2016-07-12
Support deterministic (timestamp-free) output


blendmaster:
Ideally, two `./gradlew clean shadow` runs from identical sources will produce byte-identical outputs. Currently, the outputs have timestamps set to `now()` and I think the order of files in the jar can sometimes change.

This is important for reproducible verified builds, but I personally have a different use case: I develop hadoop jobs locally, produce a shadow jar, rsync the jar to a remote machine, and run the result. Unfortunately, since the timestamps are all different, rsync's delta detection algorithm is thrown off and it essentially has to re-upload the entire jar. However, if I preprocess the jar with this script:

```java
import java.util.zip.*;
import java.nio.file.attribute.FileTime;

public class EpochJar {
public static void main(String[] args) throws Throwable {
  FileTime epoch = FileTime.fromMillis(0);
  try (
    ZipOutputStream o = new ZipOutputStream(System.out);
    ZipInputStream i = new ZipInputStream(System.in)
  ) {
    byte[] buffer = new byte[8192];
    for (ZipEntry z = i.getNextEntry(); z != null; z = i.getNextEntry()) {
      o.putNextEntry(z.setCreationTime(epoch)
                      .setLastAccessTime(epoch)
                      .setLastModifiedTime(epoch));

      int n;
      while ((n = i.read(buffer, 0, buffer.length)) > 0) {
          o.write(buffer, 0, n);
      }
    }
  }
}
}
```

Then rsync works as expected, only transferring the (small) changes. 

Supporting deterministic outputs throughout gradle is a [larger issue][0], but I think it'd be nice for shadow itself to have an option to strip timestamps from its output.

[0]: https://discuss.gradle.org/t/feature-request-reproducible-jars/15505

johnrengelman:
I feel this type of support is beyond the scope of Shadow. Especially since it's not supported by the larger Gradle framework w.r.t. JAR creation.

You can see my investigations into this same thing here: http://imperceptiblethoughts.com/2013/12/19/reproducible-multi-project-gradle-builds-part-2.html

blendmaster:
Thanks for the blog post, it was linked in couple other mailing lists as `http://imperceptiblethoughts.com/post/70592053189/reproducible-multi-project-gradle-builds-part-2` which is now a dead link.

[...]

PR #246, opened 2016-09-09
Add "zeroEntryTimestamps()" option

blendmaster:
See javadoc on the method for details on why and how. This
is basically the half-assed deterministic build from #229,
which is to say still extremely helpful when dealing with rsync.

Also update docs.gradle since doclet resolution freezes the build until the http request times out.

### DIFF
--- a/gradle/docs.gradle
+++ b/gradle/docs.gradle
@@ -1,5 +1,5 @@
 def javaApiUrl = 'http://docs.oracle.com/javase/1.7.0/docs/api'
-def groovyApiUrl = 'http://groovy.codehaus.org/gapi/'
+def groovyApiUrl = 'http://groovy-lang.org/api.html'
 
 tasks.withType(Javadoc) {
     classpath += project.configurations.shadow
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
@@ -39,6 +39,10 @@ import java.util.zip.ZipException
 
 @Slf4j
 public class ShadowCopyAction implements CopyAction {
+    /**
+     * Jan 1st, 1980. The day the zip timestamps stood still.
+     */
+    public static final int DOS_TIMESTAMP_ZERO = 315558000000L
 
     private final File zipFile
     private final ZipCompressor compressor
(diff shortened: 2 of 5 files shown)


johnrengelman:
@blendmaster thanks for the submissions. I'm still finding time in my schedule to invest into some updates for shadow. As for this, I'm considering making it a higher level feature that allows a user to register arbitrary actions against each `ZipEntry` is is produced instead of a boolean flag.
That would make it more extensible to add new capabilities.

So I'm envisioning something like adding the following method to the `ShadowJar` class:

```
interface ZipEntryAction extends Action<ZipEntry> {}

ShadowJar zipEntries(ZipEntryAction action) {
  this.zipEntryActions << action
}
```

these would be passed down to the ShadowCopy Action and then applied to each ZipEntry.

If you want to take a crack at implementing it that way, that would be cool, otherwise when I find time, i'll probably incorporate your changes that way.

blendmaster:
It seems reasonable to offer more general hooks into the final shadow zip file creation. However, I'm having trouble thinking of what anybody would do with just the ZipEntry besides zero the timestamps. Did you have anything particular in mind?

If you could somehow affect both the ZipEntry and the actual _contents_ of the entry, it might be more useful. Something like `(ZipEntry e, OutputStream original) -> OutputStream transformed`, where you can either wrap the `original` ZipOutputStream in a transform of some sort, or return it verbatim. Contrived example:

```groovy
zipEntryActions << { entry, orig ->
  if (entry.name ~= "*.java") {
    // prefix text with copyright
    def written = false
    return new FilterOutputStream(orig) {
      def write(bytes) {
        if (!written) {
          orig.write("my copyright notice\n")
          written = true
        }
        orig.write(bytes)
    }
  } else {
    return orig
  }
}
```

In any case, I can reimplement this PR on `ZipEntryAction`. I'm inclined to keep the `zeroEntryTimestamps()` option on ShadowJar as a useful shortcut, akin to `mergeServiceFiles()`.

johnrengelman:
yeah, it might end up being something like that or even extending the `Transfomer` interface to also take the `ZipEntry` as an arg...not sure. But i think implementing it this way will provide greater flexibility for moving toward something like that.

the shortcut method is fine. 

Thanks.

blendmaster:
Looking at this again, my implementation doesn't quite work with transformed (through `Transformer#modifyOutputStream(ZipOutputStream)`) files. Those use ZipOutputStream.putNextEntry themselves, avoiding the hook in ShadowCopyAction.

One implementation path I originally went down was to proxy the entire ZipOutputStream with something like javassist, allowing interception of all `putNextEntry` calls. However, I didn't want to add an extra dependency, so I went with the manual `putNextEntry` interception that seemed to (but doesn't) cover all callsites.

Given #247 and this PR, it does seem like a general way for munging the ZipOutputStream is appropriate. Let me know if you have any additional thoughts. I'll think of something to implement in the meantime.

johnrengelman:
Closing this in favor of #333

PR #247, opened 2016-09-11
Introduce keyTransformer property for PropertiesFileTransformer

marcphilipp:
The new keyTransformer property can be used to transform keys in properties files, e.g. because they contain class names about to be relocated. Its value can be set to a closure that receives the original key and returns the key name to be used in the resulting properties file.

We need this for junit-console-launcher because it uses joptsimple by @pholser which we want to relocate to a different package in order to enable users to test code that uses a different version of joptsimple. However, the joptsimple JAR contains properties files that contain keys that are class names.

### DIFF
--- a/src/docs/asciidoc/90-changes.adoc
+++ b/src/docs/asciidoc/90-changes.adoc
@@ -15,6 +15,7 @@
 * Build Shadow w/ Shadow. This will help prevent any future classpath conflicts with Gradle.
 * Replace `startShadowScripts` tasks with Gradle's built-in `CreateStartScripts` type.
 * Build with Gradle 3.1
+* https://github.com/marcphilipp[Marc Philipp] - Add `keyTransformer` property to `PropertiesFileTransformer`
 
 [discrete]
 === v1.2.4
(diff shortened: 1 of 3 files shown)


sbrannen:
Very elegant, and I'm really looking forward to seeing this in a release! :wink:

johnrengelman:
Thanks @marcphilipp for the PR. I'm still ingesting this, but off the top of my head, I'm thinking I might take this and extend it as a more 1st citizen of the processing engine so that you can register arbitrary actions to execute on each file being processed, or at least a new interface on transformer that allows you to register actions. That could help solve a few other issues that exist.

So, I'm not ignoring this, I'm just thinking about it and will be looking for time in my schedule to experiment with a few things.

marcphilipp:
@johnrengelman Any news on how to move this forward?

marcphilipp:
I've rebased my branch onto master and updated this PR.

marcphilipp:
🎉🎉🎉

johnrengelman:
@marcphilipp @sbrannen I merged this changed and pushed a snapshot build to oss.jfrog.org (2.0.0-SNAPSHOT).
You can give that shot if you'd like.

sormuras:
Thanks @johnrengelman , I will test it right away!

Btw. https://projects.ow2.org/bin/view/asm/asm-5-2-is-out with even more bug fixes. It would be great if you upgrade that dependency as well. I guess, it's only 2-times the `5.1` -> `5.2` in `shadow/gradle/dependencies.gradle`.

sbrannen:
Glad to see the merge... and the snapshot!

And an upgrade to ASM 5.2 would top it off. 😉

sormuras:
@johnrengelman See linked commit above. It works perfectly now in Java 8 and 9. Now "AssertJ" maven-shading cglib is our next show stopper: https://travis-ci.org/junit-team/junit5/builds/219055181

PR #273, opened 2017-02-04
[RFC] Implemented JAR minimization

### LINKED ISSUE #198: JAR minimization
Maven's shade plugin has the option to minimize the generated JAR, which is superb for behemoths like Guava. Is this feature a possibility?

superbobry:
This commit adds two new configuration options to the 'shadowJar' task.
When 'minimizeJar' is set the shadow JAR would only include the classes
for the project the task operates on and its dependencies. The user can
protect other classes from minimization by specifying them as entry
points.

Here's an example configuration

    shadowJar {
        minimizeJar = true
        entryPoint 'foo.Foo'
        entryPoint 'foo.Bar'
    }

N.B. imported but unused classes are not considered as dependencies. 

Closes #198

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTracker.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTracker.groovy
@@ -1,0 +1,56 @@
+package com.github.jengelman.gradle.plugins.shadow.internal
+
+import org.gradle.api.Project
+import org.gradle.api.tasks.SourceSet
+import org.vafer.jdependency.Clazz
+import org.vafer.jdependency.Clazzpath
+import org.vafer.jdependency.ClazzpathUnit
+
+/** Tracks unused classes in the project classpath. */
+class UnusedTracker {
+    private final List<String> entryPoints
+    private final List<ClazzpathUnit> projectUnits
+    private final Clazzpath cp = new Clazzpath()
+
+    private UnusedTracker(List<File> classDirs, List<String> entryPoints) {
+        this.entryPoints = entryPoints
+        projectUnits = classDirs.collect { cp.addClazzpathUnit(it) }
+    }
+
+    Set<String> findUnused() {
+        Set<Clazz> unused = cp.clazzes
+
+        for (cpu in projectUnits) {
+            unused.removeAll(cpu.clazzes)
+            unused.removeAll(cpu.transitiveDependencies)
+        }
+
+        for (entryPoint in entryPoints) {
+            Clazz clazz = cp.getClazz(entryPoint)
+            if (clazz == null) {
+                throw new RuntimeException("Entry point not found: " + className);
+            }
+
+            unused.remove(clazz)
+            unused.removeAll(clazz.transitiveDependencies)
+        }
+
+        return unused.collect { it.name }.toSet()
+    }
+
+    void addDependency(File jarOrDir) {
+        cp.addClazzpathUnit(jarOrDir)
+    }
+
+    static UnusedTracker forProject(Project project, List<String> entryPoints) {
+        final List<File> classDirs = new ArrayList<>()
+        for (SourceSet sourceSet in project.sourceSets) {
+            File classDir = sourceSet.output.classesDir
+            if (classDir.isDirectory()) {
+                classDirs.add(classDir)
+            }
+        }
+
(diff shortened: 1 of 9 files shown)


TheRealHypo:
@johnrengelman Would love to see this included. Maybe you could take a quick look, as you've been somehow absent to this repository 

@superbobry is there a way you could hand me over a version of your shadow plugin in the meantime? Have an urgent need for it ...

superbobry:
@TheRealHypo I should point out that the patch is very experimental. I've only tested it on simple examples so far, nothing real. I don't have an artifact at hand right now, but you can easily build one by [checking out](https://help.github.com/articles/checking-out-pull-requests-locally/) the PR branch and doing `./gradlew assemble`.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTracker.groovy
```diff
@@ -0,0 +1,56 @@
...
+
+        for (cpu in projectUnits) {
+            unused.removeAll(cpu.clazzes)
+            unused.removeAll(cpu.transitiveDependencies)
+        }
+
+        for (entryPoint in entryPoints) {
+            Clazz clazz = cp.getClazz(entryPoint)
+            if (clazz == null) {
+                throw new RuntimeException("Entry point not found: " + className);
```
TheRealHypo: Got an "Property not found Exception "className" on this line.

TheRealHypo:
Hey I added a Review comment on one of your files.

Also I needed to include org.apache.ant and org.vager.jDependency to build script class path.

superbobry:
I think you need to publish the review to make the comment visible.

TheRealHypo:
I tried to fix it on my own, check if the list is null etc, but after getting everything to work. The created Jar file wasn't executable, i got an error, that doesn't give me a real clue on what is missing now.

`Exception in thread "main" java.lang.ExceptionInInitializerError
	at org.codehaus.groovy.runtime.InvokerHelper.<clinit>(InvokerHelper.java:65)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.createCallConstructorSite(CallSiteArray.java:87)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallConstructor(CallSiteArray.java:60)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callConstructor(AbstractCallSite.java:235)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callConstructor(AbstractCallSite.java:239)
	at PeekCloppLinkExtract.main(PeekCloppLinkExtract.groovy:16)
Caused by: java.lang.NullPointerException
	at org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl.<init>(MetaClassRegistryImpl.java:102)
	at org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl.<init>(MetaClassRegistryImpl.java:74)
	at groovy.lang.GroovySystem.<clinit>(GroovySystem.java:36)
	... 6 more`

Because the Class "PeekCloppLinkExtract" is there, and also properly set in the Manifest.MF

TheRealHypo:
So I now know that all minimizing Jar Stuff won't work with groovy dependencies. Don't know why ...

johnrengelman:
Probably are going to need a way to specify packages that should be force included. My guess is that there's a layer of indirection happening in Groovy that the dependency library can't follow.

trustin:
Just an idea - how about using ProGuard for minimization? We (Armeria project) do this when building a shaded JAR - https://github.com/line/armeria/blob/0db6ccb1a3f2c281025e031ca219ac222a849eb3/core/build.gradle#L33

superbobry:
> Probably are going to need a way to specify packages that should be force included. My guess is that there's a layer of indirection happening in Groovy that the dependency library can't follow.

Do you know if maven-shade-plugin is able to minimize a JAR with Groovy code correctly?

Fjolnir-Dvorak:
Is this still active and/or should I help with this?

superbobry:
@Fjolnir-Dvorak I haven't looked into this for a while, feel free to take it over if you want.

skmedix:
Any progress about this? It's a really needed feature for projects with many dependencies.

johnrengelman:
This is now implemented in the latest release.

PR #302, opened 2017-06-09
Support for transitive dependencies of java-library

fkorotkov:
https://github.com/johnrengelman/shadow/pull/284 seems didn't make the new `java-library` plugin work with `shadowJar` out of the box. Seems It doesn't include dependencies transitively.

I'm still debugging and trying to figure out how to fix it on my own. In the meantime I've created a test case to show the problem. Will appreciate if @pkubowicz or @johnrengelman can hint why it can happen.

### DIFF
--- a/src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy
+++ b/src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy
@@ -424,6 +424,52 @@ class ShadowPluginSpec extends PluginSpecification {
                           'runtime.properties', 'runtimeOnly.properties'])
     }
 
+    def "include java-library configurations by default transitively"() {
+        given:
+        GradleRunner versionRunner = runner
+                .withGradleVersion('3.5')
+                .withArguments('--stacktrace')
+                .withDebug(true)
+
+        repo.module('shadow', 'api', '1.0')
+                .insertFile('api.properties', 'api')
+                .publish()
+
+        repo.module('shadow', 'implementation', '1.0')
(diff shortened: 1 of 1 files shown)


fkorotkov:
BTW `runtimeClasspath` configuration contains all `*.jar` files. But for reason `shadowJar` only copies `all.jar`. Not sure where it filters out other files.

fkorotkov:
Apparently for my Dropwizard application `runtimeClasspath` doesn't contain all from `compileClasspath` and I was getting `ClassNotFoundException` for `org/glassfish/hk2/utilities/binding/AbstractBinder`. I "fixed" it by just setting two configurations for `shadowJar` plugin:

```
shadowJar {
  configurations = [
    project.configurations.compileClasspath,
    project.configurations.runtimeClasspath
  ]
}
```

Still not sure what's going on with the test case I provided since `runtimeClasspath` does contain all jars and I've tried adding `compileClasspath` which didn't help.

johnrengelman:
I've added a test here to verify that transitive dependencies are being included - https://github.com/johnrengelman/shadow/commit/c4abfc3815e0a14af3320c2acf7269118f491fb6

I think the issue with your test is the way you are invoking `dependsOn`. It's not finding the correct method. There are 2 type signatures that are too close to each other:
```
MavenModule dependsOn(String... dependencyArtifactIds)
MavenModule dependsOn(String group, String artifactId, String version, String type = null)
```
You can fix this test by changing your invocation of this to be:
```
.dependsOn('api).dependsOn('implementation').....
```

johnrengelman:
I'm closing this, as my test shows that transitive dependencies are included. Please re-open if there is still an issue and you can provide a working example.

PR #316, opened 2017-07-28
Make build scan integration work with multi-project builds

mark-vieira:
The build scan plugin will only ever be applied to the root project. The current integration with the plugin assumes the project with the build scan plugin applied and the project with the shadow plugin applied are the same project. This will only ever be the case in a single-project build. In the case of a multi-project build these diverge.

Change the existing logic to look for the build-scan plugin on the root project, as well as include the complete task path in custom value keys as there exists the possibility that multiple `shadowJar` tasks exist within a single multi-project.


johnrengelman:
Thanks @mark-vieira.

PR #322, opened 2017-09-01
Add doc section about embedding jars

tylerbenson:
This detail tripped me up and by looking at the issues, I am not the first one.

It would be nice if there were another work around.


tylerbenson:
@johnrengelman Here's that doc update I suggested.

PR #329, opened 2017-09-30
Fix SimpleWorkResult deprecation warning in gradle 4.2

### LINKED ISSUE #326: SimpleWorkResult is deprecated in gradle 4.2
### Shadow Version
2.0.0, 2.0.1

### Gradle Version
4.2

### Expected Behavior
Clean build.

### Actual Behavior
Gradle prints a deprecation warning:
```
> Task :shadowJar
The SimpleWorkResult type has been deprecated and is scheduled to be removed in Gradle 5.0. Please use WorkResults.didWork() instead.
```

REPLicated:
Explicitly silences deprecation warning regarding SimpleWorkResult.
Since the recommended WorkResults appeared in 4.2, we can use this
workaround to both avoid this warning in newer gradle versions and still
support pre-4.2 for now.

Closes #326.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
@@ -31,6 +31,7 @@ import org.gradle.api.tasks.WorkResult
 import org.gradle.api.tasks.bundling.Zip
 import org.gradle.api.tasks.util.PatternSet
 import org.gradle.internal.UncheckedException
+import org.gradle.util.DeprecationLogger
 import org.objectweb.asm.ClassReader
 import org.objectweb.asm.ClassVisitor
 import org.objectweb.asm.ClassWriter
@@ -96,7 +97,12 @@ class ShadowCopyAction implements CopyAction {
                 )
             }
         }
-        return new SimpleWorkResult(true)
+
+        WorkResult result = null
+        DeprecationLogger.whileDisabled {
+            result = new SimpleWorkResult(true)
+        }
+        return result
     }
 
     private void processTransformers(ZipOutputStream stream) {


kenzierocks:
Perhaps it would be a good idea to add a comment explaining what is going on / when it should be replaced with the new version?

karolhor:
IMO disabling deprecation message is a little dangerous way to do it. Probable when gradle 5.0 will be released then no one will remember to change this code.

as @kenzierocks said  you could add some comments + instructions what to do in gradle 5.0 

```groovy
/**
SimpleWorkResult is deprecated and will be removed in Gradle 5.0+

When shadow will support only Gradle 4.2+ this could be changed to:
'return WorkResults.didWork(true)'
**/
```

mannodermaus:
Until the plugin's minimum required Gradle version is raised to 4.2+, why not provide a conditional to return the appropriate value instead? This feels a little cleaner to me than disabling warnings altogether for this statement:

```groovy
// WorkResults was introduced in Gradle 4.2 (TODO: Clean after raising minimum requirements)
return GradleVersion.current().compareTo(GradleVersion.version("4.2")) >= 0
    ? WorkResults.didWork(true)
    : new SimpleWorkResult(true);
```

This might even be prettified a little more with a small utility function (e.g. `isGradleVersionAbove("4.2")`).

karolhor:
Unfortunately you can't call `WorkResults.didWork(true)` in Gradle < 4.2, because this class doesn't exists before

https://github.com/gradle/gradle/blob/master/subprojects/core-api/src/main/java/org/gradle/api/tasks/WorkResults.java

mannodermaus:
That's why we'd have the Gradle version fence beforehand, no? AFAIK that's how other plugins are integrating the new WorkResult API into their flow.

REPLicated:
@mannodermaus : Have you tried executing it with gradle < 4.2? I'm sure such an approach would result in `ClassNotFoundException` when loading the containing class. Only with some reflection hacks or catching `ClassNotFoundException` you could work around that (which seems... unclean).

mannodermaus:
Yes, we're pulling in the [Gradle error-prone plugin](https://github.com/tbroyer/gradle-errorprone-plugin), which recently got a compatibility update for Gradle 4.2. They're defining the exact same [version fence](https://github.com/tbroyer/gradle-errorprone-plugin/blob/master/src/main/java/net/ltgt/gradle/errorprone/ErrorProneCompiler.java#L31) around accessing the potentially absent `WorkResults` API (as a `static final` in this case). Runtime version checks like these are also pretty common practice in domains like Android. That's why I'm convinced that we achieve compatibility across versions using a pattern like that. I can verify that an example project pulling in the error-prone Gradle plugin will execute flawlessly on both `4.1` and `4.2`. `ClassNotFound` issues like these arise when accidentally accessing absent APIs, not at the `import` level.

lfarkas:
it'd be nice to fix this annoying warning...

johnrengelman:
Fixed in https://github.com/johnrengelman/shadow/commit/f41ba29d0b838c7b80388fc9d1daf37c49e00a40

PR #333, opened 2017-10-25
Zip entry actions

johkelly:
A more general implementation of #246 , motivated by the same desire to have shadow jars that play nice with `rsync`. This implementation pushes a list of `Action<ZipEntry>` into the `ZipOutputStream` code so that all processed `ZipEntry`s can be run against them.

Arbitrary actions can be written in the build.gradle:

```
shadowJar{
    modifyZipEntries { ze ->
        println ze.hashCode()
        ze.setName("mutatedName")
        // ...
    }
}
```

And the motivating action has a sugared form:

```
shadowJar{
    zeroZipEntryTimestamps()
}
```

I'm a coworker of the requester for #246 , and lifted some of his comments and test code.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java
@@ -9,6 +9,7 @@ import com.github.jengelman.gradle.plugins.shadow.transformers.GroovyExtensionMo
 import com.github.jengelman.gradle.plugins.shadow.transformers.ServiceFileTransformer;
 import com.github.jengelman.gradle.plugins.shadow.transformers.Transformer;
 import groovy.lang.MetaClass;
+import org.apache.tools.zip.ZipEntry;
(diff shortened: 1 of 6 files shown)


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java
```diff
@@ -100,6 +103,11 @@ public ShadowJar dependencies(Action<DependencyFilter> c) {
         return this;
     }
 
+    public ShadowJar entryAction(Action<ZipEntry> action) {
```
johnrengelman: We're gonna need to make our own 1st class citizen for `ZipEntry`. We can't expose the apache classes as part of the public API for shadow because those classes are relocated during packaging.

But overall this PR looks good. I'll likely merge it this way and then figure out the stuff above.

johnrengelman:
This work will go into a 2.1.0 build since it's adding new public API.

Goooler:
I believe this is outdated, and we shouldn't apply actions for Zip entries.

PR #356, opened 2018-01-03
update inputs.file to inputs.files, to remove warning

sgnewson:
Remove warning when using the shadow plugin:
`Using TaskInputs.file() with something that doesn't resolve to a File object has been deprecated and is scheduled to be removed in Gradle 5.0. Use TaskInputs.files() instead`


ldaley:
@johnrengelman can you please put out a release with this fix. Thanks in advance!

johnrengelman:
@ldaley version 2.0.3 is available with this change.

ldaley:
@johnrengelman you're the man.

Issue #306, opened 2017-06-22
Publishing error


miclefebvre:
### Shadow Version
2.0.0

### Gradle Version
4.0

### Expected Behavior
publish to maven local

### Actual Behavior
> Failed to publish publication 'shadow' to repository 'mavenLocal'
   > Invalid publication 'shadow': POM file is invalid. Check any modifications you have made to the POM file.

It seems't that the generated publication pom file isn't working.

content of the publication pom:

<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.patate</groupId>
  <artifactId>patate-g</artifactId>
  <version>1.8.0-SNAPSHOT</version>
  <packaging>pom</packaging>
  <name>patate-g</name>
  <description/>
  <dependencies/>
  <dependencies/>
  <dependencies/>
</project>

line added to the gradle file:

publishing {
    publications {
        shadow(MavenPublication) { publication ->
            project.shadow.component(publication)
        }
    }
}

The shadowJar is working.

DPUkyle:
I'm dealing with the same situation @miclefebvre. Is your project also using `java-gradle-plugin`?

miclefebvre:
@DPUkyle yes also.

DPUkyle:
OK good. I'm working on a fix or workaround.

The problem appears to be that `com.github.jengelman.gradle.plugins.shadow.ShadowExtension#component(MavenPublication)` doesn't play nice with the `java-gradle-plugin`, which has already created a `MavenPublication` called `pluginMaven`.

In simpler terms, the `component` method always adds a new `<dependencies>` node, but should instead first check for the existing one created by `java-gradle-plugin`. I hope to create a PR today... if I don't drive myself crazy trying to figure out xml parsing in groovy first.

miclefebvre:
Amazing

rgozim:
Any update on this? The issue still exists on version 2.0.4

felipecsl:
make sure you're using the syntax below:
```
publishing {
  publications {
    shadow(MavenPublication) {
      project.shadow.component(it)
      artifact jar
    }
  }
}
```

[...]

PR #362, opened 2018-03-09
Help with 'java-gradle-plugin' without disabling `automatedPublishing`

DPUkyle:
Hi @johnrengelman , nice to meet you.

First of all, don't merge this! The test is knowingly failing.

So, my use case that I have a several collections of gradle plugins and I want to shadow their transitive dependencies to avoid buildscript classpath wonkiness (damn you, guava). I also strongly prefer to leverage the implicit publications generated by `java-gradle-plugin` when we leave the `automatedPublishing` property of GradlePluginDevelopmentExtension as its default (true). Interestingly, some of the workarounds[1] to being able to apply the `java-gradle-plugin` came from this project's build itself.

Ultimately, though, I think I've hit a wall. The way the software model rules are configured in `o.g.plugin.devel.plugins.MavenPluginPublishingRules`, there's not a way I can see to avoid adding both the un-shadowed JAR and the shadowed JAR as artifacts. I'm curious if you've run in to this and if you have any thoughts or examples on how we might work around this.

[1] Basically I have to do four things to a plugin's build to be able to use `java-gradle-plugin`:
1. Remove `gradleApi` from compile configuration
2. Remove the shadowJar's `-all` classifier
3. Call `maybeCreate('pluginMaven')` inside the PublishingContainer to avoid a conflict
4. Call component method on `ShadowExtension` (requires my fix to reuse the <dependencies> node, if it already exists. I believe this is the same symptom shown by #306 )

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -17,5 +17,4 @@ dependencies {
     testCompile 'xmlunit:xmlunit:1.3'
     testCompile 'commons-lang:commons-lang:2.6'
     testCompile 'com.google.guava:guava:17.0'
-    testCompile 'org.apache.maven:maven-model:3.0.4'
 }
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -17,4 +17,5 @@ dependencies {
     testCompile 'xmlunit:xmlunit:1.3'
     testCompile 'commons-lang:commons-lang:2.6'
     testCompile 'com.google.guava:guava:17.0'
+    testCompile 'org.apache.maven:maven-model:3.0.4'
 }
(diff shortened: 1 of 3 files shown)


johnrengelman:
Shadow doesn't create any publication. So I don't see how it's Shadow doing this. `ShadowExtension.component(MavenPublication)` takes a publication, so the author has full control over that.

johnrengelman:
This is a very old thread, but I recently had to update this project's build itself and I think this may be the workaround you are looking for - https://github.com/johnrengelman/shadow/blob/master/gradle/publish.gradle#L84-L104

johnrengelman:
Since version 1.0.0, the Gradle Plugin Publish plugin interoperates with Shadow natively (see https://plugins.gradle.org/docs/publish-plugin). Closing this issue as obsolete.

PR #363, opened 2018-03-13
Add Compatibility for Gradle 4.5 & 4.6 + Migrate to GRGIT

CDAGaming:
See https://github.com/ajoberstar/gradle-git-publish for Migration Changes

PR's Main Goal here is compatibility for Gradle 4.5.x and 4.6.x


build.gradle
```diff
@@ -87,8 +88,8 @@ idea {
     }
 }
 
-sourceCompatibility = '1.7'
-targetCompatibility = '1.7'
+sourceCompatibility = '1.8'
+targetCompatibility = '1.8'
```
johnrengelman: Can't move to JDK 8 as the minimum required version until Gradle does.

build.gradle
```diff
@@ -65,8 +66,8 @@ repositories {
 test {
     if (System.env.CI == 'true') {
         testLogging.showStandardStreams = true
-        minHeapSize "1g"
-        maxHeapSize "1g"
+        minHeapSize = "1g"
+        maxHeapSize = "1g"
```
johnrengelman: `=` not needed.

build.gradle
```diff
@@ -2,7 +2,7 @@ buildscript {
     repositories {
         jcenter()
         maven {
-            url "https://plugins.gradle.org/m2/"
+            url = "https://plugins.gradle.org/m2/"
```
johnrengelman: `=` not needed.

build.gradle
CDAGaming: IntelliJ will show this line as an error/warning (Need to find which one).

build.gradle
CDAGaming: IntelliJ will show this line as an error/warning (Need to find which one).

gpathak:
are these changes ready to go?  would be nice to get gradle 4.6 support.

CDAGaming:
These should still be all good. Will do a quick rebase to tidy things up here if this is planned to be merged soon :)

CDAGaming:
I've rebased this PR, should be all ready now :)

gradle/ghPages.gradle
```diff
@@ -1,19 +1,15 @@
-apply plugin: "org.ajoberstar.github-pages"
+apply plugin: 'org.ajoberstar.git-publish'
 
-githubPages {
+gitPublish {
     repoUri = 'https://github.com/johnrengelman/shadow.git'
-    credentials {
-        username = project.hasProperty('githubToken') ? project.githubToken : ''
-        password = ''
-    }
```
johnrengelman: With the credential configuration being removed, how does they get configured now?

build.gradle
johnrengelman: I guess my point here is that it’s not broken, so to reduce the blast radius of any change that’s coming in, please keep it to just the thing you are fixing.

johnrengelman:
@gpathak - these changes are just for the build for this plugin itself. It has no bearing on using Shadow on Gradle builds that are using 4.6.

johnrengelman:
The Gradle version compatibility is being checked here - https://github.com/johnrengelman/shadow/blob/master/src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy

gradle/ghPages.gradle
CDAGaming: Sorry i didn't get to this, but the Migration changes and how to configure it is explained in the link I gave in the original description: https://github.com/ajoberstar/gradle-git-publish

PR #390, opened 2018-07-12
Implemented JAR minimization

debanne:
(this is a follow up of https://github.com/johnrengelman/shadow/pull/273)

This commit adds two new configuration options to the 'shadowJar' task.
When 'minimize' is set the shadow JAR would only include the classes
for the project the task operates on and its dependencies. The user can
protect other classes from minimization by excluding their module.

Here's an example configuration

    shadowJar {
      minimize {
        exclude(dependency('junit:junit:.*'))
      }
    }

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -10,6 +10,7 @@ dependencies {
     compile 'commons-io:commons-io:2.5'
     compile 'org.apache.ant:ant:1.9.7'
     compile 'org.codehaus.plexus:plexus-utils:3.0.24'
+    compile 'org.vafer:jdependency:1.3'
 
     testCompile("org.spockframework:spock-core:1.0-groovy-2.4") {
         exclude module: 'groovy-all'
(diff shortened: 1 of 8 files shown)


johnrengelman:
This is great @debanne. I'm going to take a few more days to digest this and decide if it's something that I want to release in the 2.x branch or to do a major version (I believe there is some Gradle framework changes, I need to bring the plugin up to speed on as well).

Anyway, if you don't hear anything on this for a while, don't be alarmed. This looks good code-wise, I just need to figure out the release plan.

tylerbenson:
How does this work for multi project builds?  What about classes defined in `META-INF/services` files?

For example, say I have projects A -> B -> C, where A transitively depends on C.  Lets say C contains classes that are defined as services and aren't directly referenced anywhere except those service files and are loaded by A via the service loader.

In my shadow jar defined on A, I'd expect to have classes for A,B, and C all untouched.  Further I would expect any dependencies' classes used by any of those projects to also be included.

Currently, it seems to be way too aggressive and is removing many classes defined in my child projects.  The `exclude` syntax also only seems to work on library dependencies, not project dependencies.

(If you'd prefer, I can move this over to a separate issue.)

johnrengelman:
Hey @tylerbenson - can you open issues for any of the things that you have concerns about?

I didn’t implement the logic but I’m guessing it’s similar to the maven implementation? @debanne?

annunez:
Hi!

How can I exclude a jar from the project's folder inside minimize()? The path is `lib/dependency.jar`.

Thank you very much.

PR #391, opened 2018-07-27
Support AbstractArchiveTask.preserveFileTimestamps for reproducible builds

Macil:
With this PR, the [officially recommended Gradle config for reproducible builds](https://docs.gradle.org/4.8/userguide/working_with_files.html#sec:reproducible_archives) now successfully creates reproducible builds when used with shadowJar:

```
tasks.withType(AbstractArchiveTask) {
    preserveFileTimestamps = false
    reproducibleFileOrder = true
}
```

Gradle 3.4 added support for reproducible builds of archives, but shadowJar doesn't respect the preserveFileTimestamps setting despite extending AbstractArchiveTask which exposes that. This PR makes the shadowJar task handle the preserveFileTimestamps setting. (I believe the reproducibleFileOrder setting is handled by Gradle in a way that already works for shadowJar.)

Solves https://github.com/johnrengelman/shadow/issues/229 and https://github.com/johnrengelman/shadow/issues/389.

The implementation and choice of the default constant timestamp are inspired by Gradle's own ZipCopyAction class.

This does make the plugin incompatible with versions of Gradle before 3.4. I think I could change the PR to use reflection to read the preserveFileTimestamps value if backwards compatibility with those versions was considered important.

I know that the inspiration for https://github.com/johnrengelman/shadow/pull/333 was to allow a user to configure build.gradle to zero out timestamps, but I don't think this PR is redundant with it: this PR adds support to shadowJar for the standard Gradle configuration for reproducible builds without adding any new shadow APIs.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
@@ -40,6 +40,7 @@ import java.util.zip.ZipException
 
 @Slf4j
 class ShadowCopyAction implements CopyAction {
+    private static final long CONSTANT_TIME_FOR_ZIP_ENTRIES = (new GregorianCalendar(1980, 1, 1, 0, 0, 0)).getTimeInMillis()
 
     private final File zipFile
     private final ZipCompressor compressor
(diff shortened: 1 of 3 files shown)


johnrengelman:
Thanks for the PR!
This looks good and I just need to figure out the release plan for the next version before merging.

PR #393, opened 2018-08-16
Migrating Maven transformer to Gradle.

paul-nelson-baker:
I originally wrote this in its own plugin which attached itself to the shadow plugin. It's better to bundle the transformer into the shadow plugin inself instead of my own orthoginal plugin.

https://issues.apache.org/jira/browse/LOG4J2-673
https://github.com/paul-nelson-baker/gradle-shaded-log4j2-cachefile-transformer/blob/master/src/main/java/io/paulbaker/gradle/shadow/transform/Log4j2PluginsCacheFileTransformer.java
https://github.com/edwgiz/maven-shaded-log4j-transformer/blob/master/src/main/java/com/github/edwgiz/mavenShadePlugin/log4j2CacheTransformer/PluginsCacheFileTransformer.java

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/Log4j2PluginsCacheFileTransformer.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/Log4j2PluginsCacheFileTransformer.groovy
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License") you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.github.jengelman.gradle.plugins.shadow.transformers
+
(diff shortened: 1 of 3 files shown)


paul-nelson-baker:
This will actually help provide better support for AWS Lambda. Currently their instructions only zip up the project, but this actually addresses the problem that is introduced with logging in their environment.

https://docs.aws.amazon.com/lambda/latest/dg/create-deployment-pkg-zip-java.html

johnrengelman:
Looks good. Any way we can write a test for it?

paul-nelson-baker:
Sure thing! I wrote tests in my own project, but forgot to migrate them over with the plugin itself.

johnrengelman:
Looks good. Thanks!
I'm still figuring out the release plan for the next version and whether it's going to be a minor or major release.

paul-nelson-baker:
Cool! If there's anything I can do to help, please feel free to reach out!

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/Log4j2PluginsCacheFileTransformer.groovy
```diff
@@ -0,0 +1,117 @@
...
+    private final List<Relocator> relocators;
+
+    public Log4j2PluginsCacheFileTransformer() {
+        temporaryFiles = new ArrayList<>();
+        relocators = new ArrayList<>();
+    }
+
+    @Override
+    boolean canTransformResource(FileTreeElement element) {
+        return PLUGIN_CACHE_FILE == element.name
```
Goooler, line 56: This line does not work for 

```kt
@Test
fun canTransformResource() {
  assertThat(transformer.canTransformResource("")).isFalse()
  assertThat(transformer.canTransformResource(".")).isFalse()
  assertThat(transformer.canTransformResource("tmp.dat")).isFalse()
  assertThat(transformer.canTransformResource("$PLUGIN_CACHE_FILE.tmp")).isFalse()
  assertThat(transformer.canTransformResource("tmp/$PLUGIN_CACHE_FILE")).isFalse()
  assertThat(transformer.canTransformResource(PLUGIN_CACHE_FILE)).isTrue()
}
```

Seems we should use `element.relativePath.pathString` here.

PR #434, opened 2018-12-01
fix for title gets wrong

paulvi:


johnrengelman:
This isn't the way to fix this. The code in the gh-pages branch is all generated. So this needs to be configured in the `src/docs/.vuepress/config.js` file I believe.

paulvi:
Well, if it not yet solved, why to close this issue?

Yes, I know it is generated stuff, but as of moment this is quick fix, that has already drawn attention to what users see as name in browser tab.

johnrengelman:
I closed the pull request because it’s not the correct fix. I won’t merge a pull request with an incorrect fix, that I’ll have to maintain in each and every release. 

I’ll accept a PR with the correct fix or you can open an issue and I’ll fix it the next time I find time to work on it.

paulvi:
absolutely correct

PR #445, opened 2019-01-05
Add 5.0 & 5.1 compat check

3flex:


3flex:
Looks like 5.1 is incompatible - will closing for now and opening a new issue for tracking.

PR #454, opened 2019-01-29
Bugfix/protect the quotes issue 453

kyrrigle:
Proposed fix for [issue 453](https://github.com/johnrengelman/shadow/issues/453)

Note that I've switch to explicitly using bash in order to use $'string' syntax.  Not sure of another way around?


kyrrigle:
Sorry, I'm not seeing how this change is breaking the build.  When I run with --info I see errors about not being able to create files.

kyrrigle:
hadn't noticed that the $'s were escaped in the .txt file!   now it's a real proposal...

PR #456, opened 2019-02-06
Remove ShadowJar MetaClass implementation. Fix for gradle 5.2

Hillkorn:
As I tested the plugin with gradle 5.2 I got an error. Gradle was unable to create the shadowJar task because of duplicate metaClass methods. Not sure if we really have to remove it here or it's more a problem of gradle.


baron1405:
I have hit this problem in Gradle 5.2.1 and pretty much blocked unless you know of a workaround or there is a fix coming.

baron1405:
FYI @Hillkorn I cloned your branch and tried it in my Gradle 5.2.1 build. It solved the problem.

PR #459, opened 2019-02-11
Add automatic handling of Gradle API when using java-gradle-plugin

maguro:
The Java Gradle Plugin Development plugin, `java-gradle-plugin`,
automatically adds the full Gradle API to the `compile` configuration;
thus overriding a possible assignment of `gradleApi()` to the `shadow`
configuration.  Since it is never a good idea to include the Gradle API
when creating a Gradle plugin, the dependency is removed so that it is
not included in the resultant shadow jar.


maguro:
ping

maguro:
ping

PR #460, opened 2019-02-15
Fix for com.palantir.docker interoperability.

bfg:
This patch fixes a problem when `application`, `shadow` and
`com.palantir.docker` plugins are activated at the same time.

Exception:

```
An exception occurred applying plugin request [id: 'application']
> Failed to apply plugin [class
'com.github.jengelman.gradle.plugins.shadow.ShadowApplicationPlugin']
   > No such property: provider for class:
org.gradle.api.internal.artifacts.publish.ArchivePublishArtifact
```


bfg:
@johnrengelman what do you think?

Reproducer here: https://github.com/bfg/eureka-dns-server/blob/master/eureka-dns-server-standalone/build.gradle

PR #461, opened 2019-02-26
Add support for Gradle metadata

melix:
## Context

This PR is _not_ ready and only there to demonstrate the
potential of Gradle metadata. Open for discussion.

This commit adds support for Gradle metadata, available in
Gradle 5.3+. First, it makes use of the new public publication
API to create a custom software component.

This software component is registered so that when published,
we publish both the normal jars _and_ the shadow jar. When
Gradle metadata is enabled, we also publish Gradle metadata,
which will declare variants for all usages. In particular,
it makes use of the `Bundling` attribute to tell Gradle that
this particular variant is a shadow jar.

This makes it possible for consumers to _choose_ whatever
dependency they want: either the regular jar, or the shadow
one. It's worth noting that for Maven consumers, they would
have to rely on the classifier to get the all jar.


autonomousapps:
What's the status of this? I just spent several hours yesterday trying to use shadow to help bundle some dependencies into my gradle plugin before finally realizing that the `shadow.component()` method has nothing actually to do with `SoftwareComponent`s.

autonomousapps:
Somewhat related, `gradleApi()` and `localGroovy()` are no longer on the `compile` configuration. They're `api` now. I had to manually do this to strip them from my shadow jar
```kotlin
configurations["api"].dependencies.remove(dependencies.gradleApi())
configurations["api"].dependencies.remove(dependencies.localGroovy())
```

ZacSweers:
@johnrengelman is there anything further for considering this?

autonomousapps:
Was this just released? The release notes for 6.0 say this plugin now supports gradle module metadata.

johnrengelman:
It’s being per configured per the release notes but it’s untested....I don’t have the time to dig into the details. 
If the spec in this pr exhibits real world desired behavior, then we need to pull it into main.

johnrengelman:
This was included in v6.0.0 with this commit: https://github.com/johnrengelman/shadow/commit/5c572a68bebf712a99854685d6c2834d146f9924

Any feedback on real-life usage should be opened as issues.

autonomousapps:
That's great! Thanks for pinging.

autonomousapps:
I added an issue https://github.com/johnrengelman/shadow/issues/611 to provide some feedback. @melix I know this isn't your project, but you have a much deeper understanding of this stuff than I do, could you take a look?

PR #474, opened 2019-03-26
Add ManifestAppenderTransformer to support sealing packages in jars.

chrisr3:
Java's manifest attributes are implemented using a `HashMap`, which makes them useless when the attributes must be written in a particular order. E.g. for sealing packages in jars:
```
Name: com/example/
Sealed: true
```

Implement a `Transformer` which can append tags to a manifest while preserving their order.

~Note: This is a WIP - my transformer's current UI is based on Groovy's `Tuple2`. However, I'm sure that Groovy can support something better.~

The transformer's current API is:
```groovy
shadowJar {
    transform(ManifestAppenderTransformer) {
        append('Name', 'first/sealed/package/')
        append('Sealed', true)
        append('Name', 'second/sealed/package/')
        append('Sealed', true)
    }
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,6 +67,8 @@ test {
         minHeapSize "1g"
         maxHeapSize "1g"
     }
+    
+    systemProperty 'java.io.tmpdir', buildDir.absolutePath
 }
 
 jar {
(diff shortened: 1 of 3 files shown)


chrisr3:
BTW, can this plugin be installed into a local repository? I would like to test it properly on an actual project :wink:.

PR #477, opened 2019-03-29
Fixed NPE when api project dependency has a version number.

kelemen:
If an API project dependency has a version number, then the check for the dependency name was wrong (i.e., it won't end with *myproject.jar* but say *myproject-1.0.jar*). I tried to create a change which is reasonably compatible with the previous behaviour in other cases.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTracker.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTracker.groovy
@@ -2,6 +2,7 @@ package com.github.jengelman.gradle.plugins.shadow.internal
 
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.artifacts.SelfResolvingDependency
 import org.gradle.api.file.FileCollection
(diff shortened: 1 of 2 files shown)


sormuras:
I _duplicated_ this PR in #478, methinks.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTracker.groovy
```diff
@@ -58,10 +74,11 @@ class UnusedTracker {
         apiDependencies.each { dep ->
             if (dep instanceof ProjectDependency) {
                 apiJars.addAll(getApiJarsFromProject(dep.dependencyProject))
-                apiJars.add(runtimeConfiguration.find { it.name.endsWith("${dep.name}.jar") } as File)
+                addJar(runtimeConfiguration, dep, apiJars)
             } else if (dep instanceof SelfResolvingDependency) {
                 apiJars.addAll(dep.resolve())
             } else {
+                addJar(runtimeConfiguration, dep, apiJars)
```
sormuras: Why this and the next line? Ain't some duplication going on then?

kelemen: I assumed that the intent is essentially the same in both location (i.e., add only the dependency jar). Although, arguably the `isProjectDependencyFile` name makes less sense this way. If you mean the duplicate call to `addJar`, then it is difficult to not repeat something with the middle condition in the way.

kelemen:
I'm almost certain that, that issue is a completely different issue and as such it would be better to fix it in a separate PR (and only manifests here because so far the code didn't even get to that point).

Issue #465, opened 2019-03-06
Gradle 5.2, shadow 5.0.0 - sed invalid argument


DALDEI:
Please check the [User Guide](http://imperceptiblethoughts.com/shadow) before submitting "how do I do 'x'?" questions!

### Shadow Version
5.0.0
### Gradle Version
5.2.1
### Expected Behavior
should be able to run build/install/app-shadow/bin/app  someareg

### Actual Behavior

build/install/app-shadow/bin/app  somearg
sed: -e expression #1, char 23: unknown command: `''

### Gradle Build Script(s)
    id 'com.github.johnrengelman.shadow' version '5.0.0'

DALDEI:
note: if you look at the stock scripts the sed command line is slightly different
```

<     for i do printf %s\\n "$i" | sed "s/'/'\\\\''/g;1s/^/'/;\$s/\$/' \\\\/" ; done

vs

>     for i do printf %s\n "$i" | sed "s/'/'\\''/g;1s/^/'/;$s/$/' \\/" ; done

```

charleskorn:
I am also seeing a similar issue when trying to start an application with an argument.

The error I get is:

```
sed: 1: "s/'/'\''/g;1s/^/'/;/$/' \/": invalid command code '
```

This is on OS X.

The application then runs as if no arguments were passed to it.

Goooler:
Finished in #487.

PR #486, opened 2019-05-02
Add one more layer of escaping

minkenlai:
Fixes issue #465

### DIFF
--- a/src/main/resources/com/github/jengelman/gradle/plugins/shadow/internal/unixStartScript.txt
+++ b/src/main/resources/com/github/jengelman/gradle/plugins/shadow/internal/unixStartScript.txt
@@ -156,7 +156,7 @@ fi
 
 # Escape application args
 save() {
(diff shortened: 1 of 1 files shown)


src/main/resources/com/github/jengelman/gradle/plugins/shadow/internal/unixStartScript.txt
```diff
@@ -156,7 +156,7 @@ fi
 
 # Escape application args
 save() {
-    for i do printf %s\\n "\$i" | sed "s/'/'\\\\''/g;1s/^/'/;\$s/\$/' \\\\/" ; done
+    for i do printf %s\\\\n "\\$i" | sed "s/'/'\\\\\\\\''/g;1s/^/'/;\\$s/\\$/' \\\\\\\\/" ; done
```
minkenlai: For reference, gradle's script produces this line:
`for i do printf %s\\n "$i" | sed "s/'/'\\\\''/g;1s/^/'/;\$s/\$/' \\\\/" ; done`

So, besides for the `$i`, all of the escapes needed another escape.

minkenlai:
oops, buggy, will repost with correction.

PR #502, opened 2019-07-03
Use convention mapping for classifier on old gradle

### LINKED ISSUE #463: Removing the default classifier with the archiveClassifier property
### Shadow Version

5.0.0

### Gradle Version

5.2.1

### Expected Behavior

Default classifier set by `ShadowJar` task can be removed from a build script.

### Actual Behavior

I can only remove the `-all` classifier via the following code (which is deprecated):

```
tasks.named<ShadowJar>("shadowJar") {
    classifier = null
}
```

I was out of luck using the `archiveClassifier` attribute e.g. via the following code:

```
tasks.named<ShadowJar>("shadowJar") {
    archiveClassifier.set("")
}
```

The `-all` classifier is still added to the archive. Can you give an example on how to remove the `-all` classifier with the `archiveClassifier` property?

I see that the [relevant code](https://github.com/johnrengelman/shadow/blob/master/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy#L44-L51) sets the classifier via convention mapping and the new attribute `archiveClassifier`. Should the convention mapping code go into an `else` branch to only kick in when you are using an older Gradle version than 5.2?

grossws:
Fixes #463

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
@@ -41,13 +41,14 @@ class ShadowJavaPlugin implements Plugin<Project> {
         ShadowJar shadow = project.tasks.create(SHADOW_JAR_TASK_NAME, ShadowJar)
         shadow.group = SHADOW_GROUP
         shadow.description = 'Create a combined JAR of project and runtime dependencies'
-        shadow.conventionMapping.with {
-            map('classifier') {
-                'all'
-            }
-        }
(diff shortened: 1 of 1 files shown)


cvielma:
I'm not sure this fixes #463 because the classifier will always be "-all". Is there a way to get rid of the classifier?

skmedix:
https://imperceptiblethoughts.com/shadow/configuration/#configuring-output-name

cvielma:
yeah, what I was missing was that classifier is deprecated in gradle: https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:classifier but I guess it goes beyond the scope of this change.

PR #507, opened 2019-08-02
Performance optimization to perform simple (startsWith) check first b…

inez:
…efore performing more complex checks with might contain slow regexes


inez:
@johnrengelman Does this look good to you?

PR #524, opened 2019-10-29
Make shadowJar task cacheable

ghale:
These changes make the `shadowJar` task compatible with the build cache.  

The task will still be cacheable if any of the following transformers are used:

- AppendingTransformer
- XmlAppendingTransformer
- ServiceFileTransformer
- GrroovyExtensionModuleTransformer

But the other out-of-the-box transformers are not cacheable yet (if they are added, the task will be marked as un-cacheable) but I can add caching for those with a future PR.  User-defined transformers will also make the task un-cacheable unless they declare the inputs of the transformer and annotate it with `@CacheableTransformer`.

Similarly, `SimpleRelocator` has been marked as cacheable, but any custom relocators will disable caching unless the class is annotated with `@CacheableRelocator`.

We've seen fairly significant task execution times for projects with large dependency graphs, so making this task usable with the build cache could provide a decent performance improvement.

Please pay special attention to the test coverage and let me know if there are any potential edge cases not covered.  I'm happy to discuss anything if it doesn't make sense or seems wrong.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.groovy
@@ -31,27 +31,21 @@ import java.util.regex.Pattern
  * @author Mauro Talevi
  * @author John Engelman
  */
+@CacheableRelocator
 class SimpleRelocator implements Relocator {
 
-    @Input
     private final String pattern
 
-    @Input
     private final String pathPattern
 
-    @Input
     private final String shadedPattern
 
-    @Input
     private final String shadedPathPattern
 
-    @Input
     private final Set<String> includes
 
-    @Input
     private final Set<String> excludes
-
-    @Input
+    
     private final boolean rawString
 
     SimpleRelocator() {
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.groovy
@@ -20,6 +20,7 @@
 package com.github.jengelman.gradle.plugins.shadow.relocation
 
 import org.codehaus.plexus.util.SelectorUtils
+import org.gradle.api.tasks.Input
 
 import java.util.regex.Pattern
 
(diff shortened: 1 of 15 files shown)


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.groovy
```diff
@@ -188,4 +190,39 @@ class SimpleRelocator implements Relocator {
...
+
+    @Input
+    String getShadedPattern() {
+        return shadedPattern
+    }
+
+    @Input
+    String getShadedPathPattern() {
+        return shadedPathPattern
+    }
```
Goooler, line 212: @ghale

These properties are immutable and unused, can we remove them? I'm planning to remove them and migrate `includes` and `excludes` to `org.gradle.api.provider.SetProperty`.

ghale: Yeah, if these are being removed in favor of property objects, then yes, these can be removed.  They were only added in order to expose the inputs from the relocator.

ghale: To be clear, the new property objects should also be exposed as inputs (i.e. a getter with an `@Input` annotation) in order to maintain proper cacheability after these are removed.

Goooler: Can you take a review for #1079?

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/CacheableTransformer.groovy
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.Target
+
+/**
+ * Marks that a given instance of {@link Transformer} is is compatible with the Gradle build cache.
+ * In other words, it has its appropriate inputs annotated so that Gradle can consider them when
+ * determining the cache key.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@interface CacheableTransformer {
```
Goooler, line 15: @ghale 

I mark all built-in transformers cacheable, it would be great if you can take a review for #1210.

PR #535, opened 2020-01-06
Optimizations to relocation checks. Up to 3x faster

Armaxis:
Performance boost on large project (5:48 -> 2:02 min total duration).

- Changed signature of methods to accept `String` as path instead of `RelocatePathContext`/`RelocateClassContext` to prevent unnecessary allocations for builder objects
- Added extra string length checks to avoid string operations where possible
- Reordered checks in `canRelocateClass` so that expensive string replacement is the last operation
- (Main Optimization) Avoided string concatenation (which led to creating `StringBuilder` instances and expensive array copy operations) by adding check for first character being `/`: if so - perform a `startsWith` check with an offset.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/Relocator.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/Relocator.groovy
@@ -28,11 +28,11 @@ package com.github.jengelman.gradle.plugins.shadow.relocation
 interface Relocator {
     String ROLE = Relocator.class.getName()
 
-    boolean canRelocatePath(RelocatePathContext context)
+    boolean canRelocatePath(String path)
 
     String relocatePath(RelocatePathContext context)
 
-    boolean canRelocateClass(RelocateClassContext context)
+    boolean canRelocateClass(String className)
 
     String relocateClass(RelocateClassContext context)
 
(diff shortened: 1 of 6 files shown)


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/relocation/Relocator.groovy
```diff
@@ -28,11 +28,11 @@ package com.github.jengelman.gradle.plugins.shadow.relocation
 interface Relocator {
     String ROLE = Relocator.class.getName()
 
-    boolean canRelocatePath(RelocatePathContext context)
```
johnrengelman: this is a breaking change. Is it really necessary? These context objects were introduced a while ago in order to abstract the details away as I was finding that we needed to add additional data for certain features.

johnrengelman: Oh, I see, you're just removing it from some of the methods.

Armaxis: Yes, I removed context only from the methods that aren't using it. It's done to decrease the number of allocations we make and to save a bit of time on creating the builder

johnrengelman:
@Armaxis there are some conflicts now on this branch. Would you mind resolving them and then I’ll get this in as part of the 6.0 release.

Armaxis:
Sure, will do later today

Armaxis:
Sure, will do later today

PR #536, opened 2020-01-08
Added ability to include/exclude paths via real regex

Armaxis:
`SimpleRelocator` uses `plexus`'s `SelectorUtils` under the hood to perform string matching. This utility in fact supports "real" regex - as long as they are wrapped with `%regex[]` 

`SimpleRelocator` didn't had the need to utilize this functionality yet, and provided `include/exclude` methods that supported "normalized patterns" - these are eventually parsed by `SelectorUtils.matchAntPathPattern`. These patterns allow using `*` for wildcard matches, but lack the power of regex.

In our project we needed a custom regex to match certain paths, hence we added methods `includeRegex/excludeRegex`. This might be useful API for somebody else as well.
See new unit test for example usage.

Note: this PR is built on top of #535 since it uses new signature of `canRelocatePath`. If both PRs are approved, #535 is to be merged first. If only this one approved - I can rework the code to not require the new signature of method. I just really didn't wanted to have two PRs that would break things if merged together.

### DIFF
--- a/src/docs/configuration/relocation/README.md
+++ b/src/docs/configuration/relocation/README.md
@@ -32,7 +32,9 @@ Be specific as possible when configuring relocation as to avoid unintended reloc
 
 ## Filtering Relocation
 
-Specific classes or files can be `included`/`excluded` from the relocation operation if necessary.
+Specific classes or files can be `included`/`excluded` from the relocation operation if necessary. Use
+[Ant Path Matcher](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html) 
+syntax to specify matching path for your files and directories.
 
 ```groovy
 // Configuring Filtering for Relocation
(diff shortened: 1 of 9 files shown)


Armaxis:
Resolved merge conflict

johnrengelman:
@Armaxis - couldn’t we just document that if you want to use a real regex that you can use the existing include/exclude but wrap it with the `%regex[]` syntax?

It looks like that’s all this is doing, right?

Armaxis:
@johnrengelman that won't work because existing `include/exclude` call `normalizePatterns` on the patterns which replaces all `.` with `/` and removes last `/*` if there's any. This effectively breaks a regular regex. 

I can tweak `normalizePatterns` to skip an entry if it begins with `%regex` and drop having to add two more methods. This might be a cleaner way to do things, dunno why I didn't think of this before!

Armaxis:
Updated implementation to filter out Regex in `normalizePatterns`. Added section in documentation on usage and updated existing one to mention that it uses Ant Path Matcher

Armaxis:
@johnrengelman Hi! Is this PR in the 6.0.0? It's not in the release notes.

Congrats on the 6.0.0 release by the way!

johnrengelman:
Yes. It is in 6.0.0

johnrengelman:
I’ll update the release notes later tonight. Thanks for catching that.

Armaxis:
@johnrengelman awesome, going to upgrade to `6.0.0` soon then! Thanks again for all the effort put into this plugin!

PR #539, opened 2020-01-16
Fix excludes ant pattern on Windows

trask:
Fixes this use case on Windows:

```
  relocate('io.opentelemetry', 'io.opentelemetry.auto.shaded.io.opentelemetry') {
    exclude 'io.opentelemetry.auto.**'
  }
```

The fix is to explicitly pass separator (`/`) to `SelectorUtils.matchPath`. The default when it's not passed is `File.separator`, so this change should have no effect on non-Windows OS.


trask:
Thx!

Issue #541, opened 2020-01-24
Maven dependency not published for shadow dependency on project


ejjcase:
Please check the [User Guide](http://imperceptiblethoughts.com/shadow) before submitting "how do I do 'x'?" questions!

### Shadow Version
5.2.0
### Gradle Version
5.6.4
### Expected Behavior
I have a multi-project build with subprojects `foo-api` and `foo-impl`.  Both are published with `maven-publish`.  `foo-impl` uses `shadow` and has this dependency:

```kotlin
shadow(project(":foo-api"))
```

When I publish `foo-impl` I expect the .pom to declare a runtime dependency on `foo-api` with the appropriate group and version.

### Actual Behavior

The .pom does not declare any dependencies.

I can work around the problem by declaring the dependency like this instead:
```kotlin
shadow("$group:foo-api:$version")
```
but this only works if that version of `foo-api` is already published - an unsafe assumption.

Goooler:
It should work with the latest versions.

PR #543, opened 2020-01-27
Support project dependencies in the shadow configuration

ejjcase:
When  dependencies in the `shadow` configuration are project dependencies, expose them in the `.pom` in the same way as dependencies on external modules.

Use case: a multi-project build that publishes API and implementation in separate jars, where the implementation is a shadow jar and needs a dependency on the API.

Fix for #541.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
@@ -1,6 +1,7 @@
 package com.github.jengelman.gradle.plugins.shadow
 
 import org.gradle.api.Project
+import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.artifacts.SelfResolvingDependency
 import org.gradle.api.file.CopySpec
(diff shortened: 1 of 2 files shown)


vlsi:
I'm afraid the change looks strange.

Can you please clarify what is the point of manually adding a module to the generated pom?
Every time you edit pom.xml it creates a discrepancy between the published `pom.xml` and `Gradle Metadata`.

So if you want something to appear (or disappear) in pom.xml, it should be implemented with altering the relevant Gradle configurations rather than edits to the `pom.xml` string. That would ensure pom.xml does not contradict metadata file.

ejjcase:
Hmm, it should probably use `SoftwareComponentFactory` to create a new component to publish.  However, `SoftwareComponentFactory` (and Gradle Module Metadata) didn't exist until Gradle 5.3, and this code was written to support older Gradle versions.

But `shadow` now requires Gradle 6, so it could be changed now.

This seems like a new bug to me, introduced by the availability of a new form of metadata.

johnrengelman:
Yeah, I'm trying to navigate the various patterns here. I previously had create my own "component" using the `shadow.component` method because `SoftwareComponent` wasn't public API.

PR #549, opened 2020-02-26
Upgrade to ASM 7.3.1 to support JDK 15

vyazelenko:
@johnrengelman Any estimate on when the next version of the Shadow plugin will be released?

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -5,8 +5,8 @@ dependencies {
     shadow 'org.codehaus.groovy:groovy-backports-compat23:2.4.15'
 
     implementation 'org.jdom:jdom2:2.0.6'
-    implementation 'org.ow2.asm:asm:7.2'
-    implementation 'org.ow2.asm:asm-commons:7.2'
+    implementation 'org.ow2.asm:asm:7.3.1'
+    implementation 'org.ow2.asm:asm-commons:7.3.1'
     implementation 'commons-io:commons-io:2.5'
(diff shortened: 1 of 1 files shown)


VoxelGod:
Unfortunately, a JDK 14 and 15 preview feature (records) requires ASM 8.
```
* What went wrong:
Execution failed for task ':shadowJar'.
> Records requires ASM8
```

johnrengelman:
6.0.0 is using ASM 8.0.1 - https://github.com/johnrengelman/shadow/blob/master/gradle/dependencies.gradle#L8

VoxelGod:
@johnrengelman One of my colleagues inspected the stacktrace (https://hastebin.com/egemarucaw.txt) and found ASM7 references in jdependency to be responsible:
https://github.com/tcurdt/jdependency/blob/c2e760c533225d313c5ef8e1d1f14840591bf431/src/main/java/org/vafer/jdependency/asm/DependenciesClassAdapter.java#L128

I can notify them of the issue, then jdependency would need to be updated in ShadowJar as well.

johnrengelman:
Looks like jdependency 2.4.0 uses ASM 8

VoxelGod:
Is it possible this dependency update could be targeted for a minor release in the future? For now, I'll exclude records from my project.

bridgettewp:
Encountering the same issue with JDK 14 preview features. Records and instanceof pattern matching throw a TaskExecutionException. Relocation and minimization do work now, though.

PR #555, opened 2020-03-18
Allow package exclusion for ConfigureShadowRelocation

ward-eric:
We've got a use case where we'd like to relocate all of our dependencies with 2 exceptions:

1. We have internal dependencies that contain a package that also exists in the project that is the target of shadowJar.  This causes all of the classes in that package to be relocated, which we don't want.
2. We have another library that is a thin wrapper around a 3rd party library.  We'd like to relocate all of the dependencies of this wrapper except for the specific 3rd party library.

Instead of using `relocate` and having to explicitly list all of the packages to include we'd rather use `ConfigureShadowRelocation` and explicitly exclude the 2 packages we don't want relocated.


heychazza:
I'm looking for this exact thing! Any update with it?

zinking:
can this be revived and merged ?
@ward-eric

samotoo:
Any chance to merge it to main branch?

Goooler:
Closed due to #1094.

Issue #565, opened 2020-04-15
Support a full shrinking mode in for UnusedTracker using R8


netomi:
This is a suggestion for an improvement to an existing feature of this awesome gradle plugin.

The current minimize feature is relatively simple and is not able to shrink all unused classes from 3rd party libraries. For my own use-case this was not sufficient, having a full shrinking support together with relocation re-use of a lot of classes not available as stand-alone jars / classes would be much easier (that was actually my use-case, I wanted to use single classes from large jars without copying the content).

The shadowJar plugin has already all the features that you need for this purpose, it just lacks support for a full shrinker like ProGuard or R8. ProGuard is unfortunately GPL so I decided to provide a version of the UnusedTracker class that uses R8 for doing the dirty work.

Please take a look at my pull request. I tested it with various projects (even with the R8 project itself) and it works pretty fine. I am open to suggestions and improvements. The current integration was a quick and dirty proof of concept.

I would be happy if the feature request would be accepted.

The pull request currently uses the latest dev version of R8 which is not yet available on the normal google maven repo.

netomi:
I have added a demonstration of this pull request to this branch of my project:

https://github.com/netomi/uom/tree/shadow-dependencies

It pulls 1 class from hibernate-validator and uses 1 method from guava:

 * ConcurrentReferenceHashMap
 * Suppliers.memoize

with the original minimize feature many more classes would be shadowed, like annotation classes for code in guava ....

With the use of R8 this can be greatly reduced, only necessary classes are shadowed which would allow me to use this functionality rather than copying or rewriting the code when you dont want to pull in such a big dependency.

gildor:
@netomi Maybe this issue and linked PR would be accepted if use a bit different approach: Extract UnusedTrackerWithR8 to own dependency, contribute to ShadowCopyAction.groovy only parts required to provide custom tracker, add DSL to `shadow` extension which allows specifying custom UnusedTracker (it may be as simple as providing a factory for UnusedTrackerWithR8 and pass all configs as arguments to UnusedTrackerWithR8)

I believe this feature has a great potential, especially if it would be possible to configure r8 with config file.

If all required APIs on ShadowPlugin would be in place and it would be possible to provide custom UnusedTracker it makes it easier to try it and do not depend on ShadowPlugin release cycle

netomi:
[...]

PR #566, opened 2020-04-15
Add version of UnusedTracker that supports full shrinking using R8.

netomi:
Pull request for #565 .

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/Log4j2PluginsCacheFileTransformer.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/Log4j2PluginsCacheFileTransformer.groovy
@@ -41,6 +41,7 @@ import static org.apache.logging.log4j.core.config.plugins.processor.PluginProce
  * @see <a href="https://github.com/edwgiz/maven-shaded-log4j-transformer">edwgiz/maven-shaded-log4j-transformer</a>
  * @see <a href="https://github.com/edwgiz/maven-shaded-log4j-transformer/blob/master/src/main/java/com/github/edwgiz/mavenShadePlugin/log4j2CacheTransformer/PluginsCacheFileTransformer.java">PluginsCacheFileTransformer.java</a>
  */
+@CacheableTransformer
 class Log4j2PluginsCacheFileTransformer implements Transformer {
 
     private final List<File> temporaryFiles


netomi:
There are some test failures wrt to exclude settings. I have to look into them, so far I did not take these settings into account.

netomi:
I have fixed the test failures. There is one remaining but this one can not be easily solved when using a full shrinker. The test setup is as follows:

an unused class in an api dependency is implementing an interface from a transitive dependency. This interface is never used or accessed, thus the interface is removed, leaving no trace of it. The test want to ensure that the interface is not remove (LibEntity). Cases like that are normally not handled by a shrinker, it usually can safely remove such interfaces. If the test would be slightly adapted to a more real world example (LibEntity would be used somehow in the otherwise UnusedEntity), the interface would not be removed and the test would succeed.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
```diff
@@ -0,0 +1,265 @@
...
+                lastString = s
+            }
+        })
+
+        List<String> proguardConfig = getKeepRules()
+        proguardConfig.add("-dontoptimize")
+        proguardConfig.add("-dontobfuscate")
+        proguardConfig.add("-ignorewarnings")
+
+        builder.addProguardConfiguration(proguardConfig, Origin.unknown())
```
sgjesse, line 123: I will suggest that you use debug mode as well `builder.setMode(CompilationMode.DEBUG)` and not rely on `-dontoptimize` to turn off structural optimizations.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
```diff
@@ -0,0 +1,265 @@
...
+        // disable everything, we just want to get a list of
+        // all project classes.
+        List<String> configs = new ArrayList<>()
+        configs.add("-dontshrink")
+        configs.add("-dontoptimize")
+        configs.add("-dontobfuscate")
+        configs.add("-ignorewarnings")
+        configs.add("-dontwarn")
+
+        builder.addProguardConfiguration(configs, Origin.unknown())
```
sgjesse, line 183: Debug mode here as well.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
```diff
@@ -0,0 +1,265 @@
...
+        // add all dependencies as library jars to avoid warnings
+        for (File dep : dependencies) {
+            Path path = Paths.get(dep.getAbsolutePath())
+            builder.addLibraryFiles(path)
+        }
+
+        // disable everything, we just want to get a list of
+        // all project classes.
+        List<String> configs = new ArrayList<>()
+        configs.add("-dontshrink")
```
sgjesse, line 177: With `-dontshrink` isn't this just going to generate the list of the classes in the jar to be shadowed?

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
sgjesse: Thinking about this again this should not use debug mode. Shouldn't `-dontoptimize` and `-dontobfuscate` be removed here? There are `-keep` rules for all classes in the shadowed jar, and the dependent classes could be renamed with a `-repackageclasses` or `-flattenpackagehierachy`. That will of course require collecting (multiple) mapping files.

build.gradle
```diff
@@ -60,6 +60,13 @@ apply from: file('gradle/dependencies.gradle')
 
 repositories {
     jcenter()
+
+    // for r8 dependency
+    repositories {
```
ZacSweers, line 65: For what it's worth, you can also get stable releases from `google()`, but they're tied to AGP updates

gradle/dependencies.gradle
```diff
@@ -15,6 +15,8 @@ dependencies {
         exclude group: 'org.ow2.asm'
     }
 
+    implementation 'com.android.tools:r8:2.1.16-dev'
```
ZacSweers, line 18: Wouldn't you want a non-dev build?

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
netomi: the repackaging is performed by the shadow jar plugin itself using its remapping functionality which is more powerful than the -repackageclasses. I basically want to disable any feature of R8 apart from shrinking. If you have better ideas to achieve that they are of course welcome.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
netomi: yes thats the idea. I did not want to explode any jar file that is declared as input and collect classes, but rather feed R8 with the input and collect the classnames from the output. This is not necessary and could be improved.

build.gradle
netomi: the latest one available from google() was 1.6.84 which does not include some classes that I need, mainly the JdkClassFileProvider.

I am happy to use published releases once they are available for my use-case.

gradle/dependencies.gradle
netomi: yes, this is mainly a proof-of-concept, I did a lot of testing lately with different versions of r8, maybe the latest release 1.6.84 already includes all things that I need, I will check.

sgjesse: Right now we only release non-dev builds to Google Maven when a stable release of Android Studio is released. Which is why there is no 2.0 or 2.1 versions there yet.

netomi: I checked, the main reason to use a 2.x release is the JdkClassFileProvider. It is needed to support JDK 9+.

netomi: changed to the latest non-dev release 2.0.61.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
netomi: changed.

covers1624:
Is there any update on this PR? I am interested in using this functionality.

build.gradle
gildor: 2.0.99 available on maven.google.com now

bowsersenior:
It would be nice if we could have some feedback on this PR. It seems very useful from my team's needs.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java
```diff
@@ -98,7 +98,7 @@ public InheritManifest getManifest() {
     @Override
     protected CopyAction createCopyAction() {
         DocumentationRegistry documentationRegistry = getServices().get(DocumentationRegistry.class);
-        final UnusedTracker unusedTracker = minimizeJar ? UnusedTracker.forProject(getProject(), configurations, dependencyFilterForMinimize) : null;
+        final UnusedTracker unusedTracker = minimizeJar ? UnusedTrackerWithR8.forProject(getProject(), configurations, dependencyFilterForMinimize) : null;
```
johnrengelman, line 101: That choice of using the standard tracker vs the r8 tracker needs to be configurable by the user. 
A couple of options here:
1. Add an `useR8 = true` option that works in conjunction with `minimizeJar` to configure the implementation.
2. Detect the presence of the R8 classes on the classpath and then use the R8 implementation if present.

gradle/dependencies.gradle
```diff
@@ -15,6 +15,8 @@ dependencies {
         exclude group: 'org.ow2.asm'
     }
 
+    implementation 'com.android.tools:r8:2.0.61'
```
johnrengelman, line 18: I think we should make this `compileOnly` and force the user to add their only r8 dependency so we don't have to maintain the compatibility. This should happened regardless of the extension implementation in the comment below.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTracker.groovy
```diff
@@ -12,18 +12,32 @@ import org.vafer.jdependency.Clazz
...
+    protected final Clazzpath cp = new Clazzpath()
 
-    private UnusedTracker(List<File> classDirs, FileCollection classJars, FileCollection toMinimize) {
+    protected UnusedTracker(List<File> classDirs, FileCollection classJars, FileCollection toMinimize) {
         this.toMinimize = toMinimize
         projectUnits = classDirs.collect { cp.addClazzpathUnit(it) }
         projectUnits.addAll(classJars.collect { cp.addClazzpathUnit(it) })
     }
 
+    boolean performsFullShrinking() {
```
johnrengelman, line 29: None of these methods are necessary in this class since they aren't used by it. If they are only needed by the R8 implementation, then they should reside in that class.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * An implementation of an UnusedTracker using R8.
+ * <p>
+ * This class is written mainly in Java syntax and uses @CompileStatic deliberately
+ * to avoid groovy madness especially in combination with R8.
+ * <p>
+ * It is an extension of the existing UnusedTracker to make it easy to exchange them,
+ * if the pull request is accepted, a common base interface should be extracted.
+ */
+class UnusedTrackerWithR8 extends UnusedTracker {
```
johnrengelman, line 25: We could just write this in pure Java in the `src/main/java` directory if this is a concern.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/UnusedTrackerWithR8.groovy
```diff
@@ -0,0 +1,276 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    @CompileStatic
+    static void addJDKLibrary(BaseCommand.Builder builder) {
+        String JAVA_HOME = System.getenv("JAVA_HOME")
+        builder.addLibraryResourceProvider(JdkClassFileProvider.fromJdkHome(Paths.get(JAVA_HOME)))
```
johnrengelman, line 252: I'm not sure we can/should rely on this. Perhaps this needs to be exposed to the user directly? We could maybe figure out a default setting based on the configuration of the `compileMain` task or somethign.

src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy
```diff
@@ -495,12 +495,12 @@ class ShadowPluginSpec extends PluginSpecification {
 
         file('lib/src/main/java/lib/LibEntity.java') << """
             package lib;
-            public interface LibEntity {}
+            public interface LibEntity { void simple(); }
```
johnrengelman, line 498: we should follow appropriate Java styling and put this on a new line for ease of reading. Same for all the changes in this file.

johnrengelman (changes requested):

nicolabeghin:
Any update on this PR? seems really promising! thanks

netomi:
Somehow I missed that changes have been requested to this PR which make total sense and I would be willing to work on them.
In the meantime I worked on my own bytecode manipulation library including a shrinker: https://github.com/netomi/bat and I created a fork of  the shadow plugin to use it for relocation and shrinking (still to be done): https://github.com/netomi/gradle-shadow

It takes the current shadow plugin, converts it to kotlin only and uses my bat library for the byte code stuff. Its still very early work,  but I will come back to this PR and make it clean to be included. @johnrengelman

netomi:
Reopen the PR as it was closed by accident.

netomi:
When I sync'd my PR branch with latest master and reworked the PR the PR was closed as a consequence. I am not sure how to fix that, but I created a new PR at #796 , its a pity that the comments and suggestions are lost somehow, but I hope that suffices.

PR #591, opened 2020-07-23
Fixes to allow Configuration Caching in Gradle 6.6+

timyates:
We (myself and @britter) believe this PR allows the Shadow Plugin to be used
in a Gradle 6.6+ build with --configuration-cache enabled.

We unfortunately did not get round to fixing the case where `minimize()` is
used, so it does not work if that is called in the build.

We're basically looking for suggestions on what we can do to improve it,
or whether it is good enough to go in as-is

Cheers

Tim (and Benedikt)

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
@@ -54,12 +54,10 @@ class ShadowApplicationPlugin implements Plugin<Project> {
     }
 
     protected void configureJarMainClass(Project project) {
-        ApplicationPluginConvention pluginConvention = (
-                ApplicationPluginConvention) project.convention.plugins.application
-
-        jar.inputs.property('mainClassName', { pluginConvention.mainClassName })
+        def classNameProvider = project.provider { project.convention.plugins.application.mainClassName }
+        jar.inputs.property('mainClassName', classNameProvider)
         jar.doFirst {
-            manifest.attributes 'Main-Class': pluginConvention.mainClassName
+            manifest.attributes 'Main-Class': classNameProvider.get()
         }
     }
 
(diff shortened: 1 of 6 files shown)


timyates:
(upcoming configuration cache feature is described here https://docs.gradle.org/release-candidate/userguide/configuration_cache.html)

eskatos:
@timyates, @britter, any chance you could add a failing test for `minimize()`? I think this could help moving this forward

britter:
@eskatos I think you only need to call `minimize()` in the `ShadowJar` configuration. Haven't checked it, but this test case should fail:

```groovy
    def "configuration caching supports minimize"() {
        given:
        buildFile << """
            shadowJar {
               minimize()
            }
        """.stripIndent()

        when:
        runner.withArguments('--configuration-cache', 'shadowJar').build()
        output.delete()
        def result = runner.withArguments('--configuration-cache', 'shadowJar').build()

        then:
        contains(output, ['a.properties', 'b.properties'])

        and:
        doesNotContain(output, ['a2.properties'])
        result.output.contains("Reusing configuration cache.")
    }
```

eskatos:
With the latest changes above, the `minimize {}` use case now works with the configuration cache enabled.

@johnrengelman this should be ready for review, could you please take a look?

BTW it looks like the circleci build is broken
https://app.circleci.com/pipelines/github/johnrengelman/shadow/42/workflows/1b019a05-d7a1-47e5-aa42-3a45781b228e/jobs/497

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
```diff
@@ -54,12 +54,10 @@ class ShadowApplicationPlugin implements Plugin<Project> {
...
     protected void configureJarMainClass(Project project) {
-        ApplicationPluginConvention pluginConvention = (
-                ApplicationPluginConvention) project.convention.plugins.application
-
-        jar.inputs.property('mainClassName', { pluginConvention.mainClassName })
+        def classNameProvider = project.provider { project.convention.plugins.application.mainClassName }
+        jar.inputs.property('mainClassName', classNameProvider)
         jar.doFirst {
-            manifest.attributes 'Main-Class': pluginConvention.mainClassName
+            manifest.attributes 'Main-Class': classNameProvider.get()
```
vlsi, line 60: manifest.attributes support providers: https://twitter.com/VladimirSitnikv/status/1260952932981313544

PR #620, opened 2020-12-09
Add a whitespace trimming mode to ManifestAppenderTransformer

jeremyk-91:
Hi there,

We use the ManifestAppenderTransformer in a Gradle plugin that wraps the Shadow plugin: https://github.com/palantir/gradle-shadow-jar. 

We've tried to use the ManifestAppenderTransformer to add tags to an existing manifest file. However, existing manifest files often have a blank line at the end, and so the current transformer if used to extend such a file will leave a blank line between the existing content and any added tags. In other words, if we tried to add the tag `bar: 1234` to a manifest with a tag `foo: 123` _and a blank line at the end_, we get

```
foo: 123

bar: 1234
```

This PR extends the transformer with a mode where it trims whitespace, so that tags are added to the last section of the manifest (as opposed to beginning a new section). This would allow using this transformer to add tags to the main section of an existing manifest (which is what we are trying to do with the plugin when shading a multi-release jar). With this mode you would obtain

```
foo: 123
bar: 1234
```

This is strictly opt-in, so existing users should not be affected.

@iamdanfox for SA

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ManifestAppenderTransformer.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ManifestAppenderTransformer.groovy
@@ -72,7 +72,8 @@ class ManifestAppenderTransformer implements Transformer {
         ZipEntry entry = new ZipEntry(MANIFEST_NAME)
         entry.time = TransformerContext.getEntryTimestamp(preserveFileTimestamps, entry.time)
         os.putNextEntry(entry)
-        os.write(manifestContents)
+        os.write(trimTrailingWhitespace(manifestContents))
+        os.write(EOL)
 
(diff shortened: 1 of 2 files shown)


johnrengelman:
Is there a reason why wouldn’t want to simply always do this? Does it really need to be configurable?

jeremyk-91:
Ah, I was concerned (maybe too concerned?) about breaking users who might rely on existing whitespace, though I can't think of a use case for that offhand. If you think this will be fine, I'll change it to not be configurable.

johnrengelman:
I think maybe level it configurable but default it to enabled. So if someone _does_ have a problem, they could disable it.

Goooler:
@jeremyk-91 Would you mind porting this change to Kotlin for the latest codebase?

Goooler:
I'll port the PR to Kotlin if someone requests this again.

PR #622, opened 2020-12-10
Introduce CleanProperties

simPod:
PropertiesFileTransformer uses `java.utils.Properties` internally as a storage.

`java.utils.Properties` `store0()` contains `bw.write("#" + new Date().toString());` that prepends current timestamp before any content (after comments).

 This effectively breaks reproducible builds that use PropertiesFileTransformer because every new build has different timestamp in transformed files.

 CleanProperties implementation is introduced in order to remove prepended timestamp when creating output stream.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/CleanProperties.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/CleanProperties.groovy
@@ -1,0 +1,32 @@
+/*
+ * Source https://stackoverflow.com/a/39043903/519333
+ */
+package com.github.jengelman.gradle.plugins.shadow.internal
+
+class CleanProperties extends Properties {
+    private static class StripFirstLineStream extends FilterOutputStream {
+
+        private boolean firstLineSeen = false
+
+        StripFirstLineStream(final OutputStream out) {
+            super(out)
+        }
+
+        @Override
+        void write(final int b) throws IOException {
+            if (firstLineSeen) {
+                super.write(b);
+            } else if (b == '\n') {
+                super.write(b);
+
+                firstLineSeen = true;
+            }
+        }
+
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/CleanProperties.groovy
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.jengelman.gradle.plugins.shadow.internal
+
+class CleanProperties extends Properties {
+    private static class StripFirstLineStream extends FilterOutputStream {
```
johnrengelman, line 19: Was this copied from another open source project? The copyright in the license header says 2012.

If this was taken from another project, please include a reference to where it was obtained from in a comment on the class.

simPod: Ah, no. I first copy pasted properties and this collapsed in my IDE so did not notice the comment. It was actually taken from SO, added link to the post.

simPod:
@johnrengelman can I do anything to move this forward?

grongor:
@johnrengelman please, can we do anything to move this forward? We are deploying several java applications in one pipeline, and it would be really nice if we could skip the ones that didn't change. And we can't do that now since the hash of the jar always changes...

crafton:
@johnrengelman will this change be moving forward?

simPod:
Yet another rebase in few years. Fresh master.

k-brooks:
As described, this issue is still present.

PR #646, opened 2021-03-12
Add input/output annotations to transformers

wolfs:
so that the transformer remain compatible with Gradle 7.0.
In Gradle 7.0 missing annotations on input properties will fail the build. Input properties include properties on `@Nested` inputs, which includes all the `Transformers`.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheNoticeResourceTransformer.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheNoticeResourceTransformer.groovy
@@ -23,6 +23,9 @@ import org.apache.tools.zip.ZipEntry
 import org.apache.tools.zip.ZipOutputStream
 import org.codehaus.plexus.util.StringUtils
 import org.gradle.api.file.FileTreeElement
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Internal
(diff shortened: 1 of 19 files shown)


melix (approved):
LGTM :+1:

helfper:
I hadn't seen this before. Interesting that we came up with very similar solutions independently (#647). I believe you beat me to it with the time I spent figuring out why the test was not working for `ManifestResourceTransformer` and then fixing it upstream in https://github.com/apache/ant.

I think the main differences are:
* I preferred to make fields `private` instead of annotating them with `@Internal`
* Making `Transformer` implement `Named` shows the name of the transformer in the warnings
* Asserting that there's no deprecation message in the output line by line instead of the whole output makes it easier to find the offending line when it fails

wolfs:
@helfper I looked at your PR and it looks good! I left some comments regarding the annotations. I still would like to salvage the changes I did here to assert that no test has any deprecation warnings. I'll extract this into another PR and then close this one. Or do you prefer that I base this PR on top of yours?

wolfs:
@helfper I extracted the assertion code to https://github.com/johnrengelman/shadow/pull/654. There I also assert line-wise like you did.

johnrengelman:
Closing in favor of https://github.com/johnrengelman/shadow/pull/647 and https://github.com/johnrengelman/shadow/pull/654 per the conversation above

PR #569, opened 2020-04-30
Eliminate serviceEntries deprecation warning by making it private

jeffalder:
Fixes #530 

I ran a test between HEAD and this:

- HEAD (43a8d4e8ea5d07a29d1d012c16fd4e34f646f87c):
```
% ./gradlew --warning-mode=all shadowJar
Starting a Gradle Daemon (subsequent builds will be faster)

> Task :shadowJar
Property 'transformers.$0.serviceEntries' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0. See https://docs.gradle.org/6.3/userguide/more_about_tasks.html#sec:up_to_date_checks for more details.

BUILD SUCCESSFUL in 30s
2 actionable tasks: 2 executed
```

- this change:
```
% ./gradlew --warning-mode=all shadowJar

BUILD SUCCESSFUL in 6s
2 actionable tasks: 1 executed, 1 up-to-date
```

I also verified that the contents of the `META-INF/services` files were the same. I also ran an incremental build (first `clean shadowJar` then `shadowJar`) and the result was the same.

Issue #530, opened 2019-11-26
Gradle 6 deprecation warnings


boris-petrov:
### Shadow Version
5.2.0

### Gradle Version
6.0.1

### Expected Behavior
No warnings

### Actual Behavior
```
Property 'transformers.$0.serviceEntries' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.                                                                                                                                             
Property 'transformers.$2.data' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.                                                                                                                                                       
Property 'transformers.$3.doc' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.                                                                                                                                                        
Property 'transformers.$3.ignoreDtd' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
```

### Gradle Build Script(s)
```gradle
task foo(type: ShadowJar) {
	from sourceSets.some.output

	configurations = [project.configurations.someRuntimeClasspath]

	archiveBaseName = 'some'
	zip64 = true

	mergeServiceFiles()
	mergeGroovyExtensionModules()
	transform(AppendingTransformer) {
		resource = 'META-INF/extensions.idx'
	}
	transform(XmlAppendingTransformer) {
		resource = 'META-INF/vfs-providers.xml'
	}

	manifest.attributes('Main-Class': 'com.company.Some')

	exclude 'META-INF/*.RSA', 'META-INF/*.SF','META-INF/*.DSA'

	duplicatesStrategy = DuplicatesStrategy.FAIL
}
```

iainmcgin:
@johnrengelman I still see a warning related to the `serviceEntries` property in [ServiceFileTransformer](https://github.com/johnrengelman/shadow/blob/a7f0908521029a848b5de2e2a131af146bd213cd/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ServiceFileTransformer.groovy#L54) when running a build with 5.2.0 of the plugin:

```
Property 'transformers.$0.serviceEntries' is not annotated with an input or output annotation.
```

I don't know much about Gradle internals, but from digging around the documentation I think this property should be annotated as `@Input`.

xalvarez:
@iainmcgin bear in mind that v5.2.0 was released last November so it doesn't include the changes above.

[...]

PR #647, opened 2021-03-13
Fix transformers deprecation warnings

### LINKED ISSUE #601: Gradle 6.6.1 still reports shadowJar has a forward-incompatibility with Gradle 7.0
This is to report that, despite all the good work to address #530, there seems to be a leftover `Property 'transformers.$1.data' is not annotated with an input or output annotation`.

### Shadow Version

```groovy
id 'com.github.johnrengelman.shadow' version '6.1.0'
```

### Gradle Version

```
------------------------------------------------------------
Gradle 6.6.1
------------------------------------------------------------

Build time:   2020-08-25 16:29:12 UTC
Revision:     f2d1fb54a951d8b11d25748e4711bec8d128d7e3

Kotlin:       1.3.72
Groovy:       2.5.12
Ant:          Apache Ant(TM) version 1.10.8 compiled on May 10 2020
JVM:          1.8.0_265 (Private Build 25.265-b01)
OS:           Linux 5.4.0-47-generic amd64
```

### Expected Behavior

No forward-incompatibility with Gradle 7.0.

### Actual Behavior

```sh
$ ./gradlew --debug assemble
...
[LIFECYCLE] [class org.gradle.internal.buildevents.TaskExecutionLogger] > Task :my:shadowJar UP-TO-DATE
[DEBUG] [org.gradle.internal.operations.DefaultBuildOperationExecutor] Build operation 'Task :my:shadowJar' started
[DEBUG] [org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter] Putting task artifact state for task ':my:shadowJar' into context took 0.0 secs.
[WARN] [org.gradle.internal.featurelifecycle.LoggingDeprecatedFeatureHandler] Property 'transformers.$1.data' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0. See https://docs.gradle.org/6.6.1/userguide/more_about_tasks.html#sec:up_to_date_checks for more details.
[DEBUG] [org.gradle.internal.operations.DefaultBuildOperationExecutor] Build operation 'Snapshot inputs and outputs before executing task ':my:shadowJar'' started
...
```

### Gradle Build Script(s)

```groovy
plugins {
...
    id 'com.github.johnrengelman.shadow' version '6.1.0' apply false
...
}
...
project(':my') {
...
    apply plugin: 'com.github.johnrengelman.shadow'
...
    shadowJar {
        mergeServiceFiles()
        append('my.conf')
    }
...
}
```

helfper:
This PR annotates transformers' configuration fields as inputs to solve the deprecation warnings.

Closes #601 
Fixes the remaining warnings in #530 (only the first one was addressed in #569)

Caveats:
* I've marked `PropertiesFileTransformer.keyTransformer` as internal, while I believe it should also be an input. Just annotating it with `@Input` will cause Gradle to eagerly calculate its value (passing `null` as parameter) which is not desirable. The solution may be wrapping the closure with `Property<>`, but I'm not totally sure. At least one other plugin in a similar situation seem to have gone with internal as well (https://github.com/ben-manes/gradle-versions-plugin/blob/master/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy#L76)
* Had to comment out the test for `ManifestResourceTransformer`, because as far as I can see, that transformer has been broken since `JarOutputStream` was replaced by `ZipOutputStream` in version 0.9.0/1.0.0. This is independent of the changes in this PR and should be solved when this is released upstream: https://github.com/apache/ant/pull/145

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheNoticeResourceTransformer.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheNoticeResourceTransformer.groovy
@@ -23,6 +23,8 @@ import org.apache.tools.zip.ZipEntry
 import org.apache.tools.zip.ZipOutputStream
 import org.codehaus.plexus.util.StringUtils
 import org.gradle.api.file.FileTreeElement
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Optional
 
 import java.text.SimpleDateFormat
 
@@ -34,34 +36,46 @@ import java.text.SimpleDateFormat
  * @author John Engelman
  */
 class ApacheNoticeResourceTransformer implements Transformer {
-    Set<String> entries = new LinkedHashSet<String>()
 
-    Map<String, Set<String>> organizationEntries = new LinkedHashMap<String, Set<String>>()
+    private Set<String> entries = new LinkedHashSet<String>()
 
+    private Map<String, Set<String>> organizationEntries = new LinkedHashMap<String, Set<String>>()
+
+    @Input
     String projectName = "" // MSHADE-101 :: NullPointerException when projectName is missing
 
+    @Input
     boolean addHeader = true
 
+    @Input
     String preamble1 = "// ------------------------------------------------------------------\n" +
             "// NOTICE file corresponding to the section 4d of The Apache License,\n" +
             "// Version 2.0, in this case for "
 
+    @Input
     String preamble2 = "\n// ------------------------------------------------------------------\n"
 
+    @Input
     String preamble3 = "This product includes software developed at\n"
 
-    //defaults overridable via config in pom
+    @Input
     String organizationName = "The Apache Software Foundation"
 
+    @Input
     String organizationURL = "http://www.apache.org/"
 
+    @Input
     String inceptionYear = "2006"
 
+    @Optional
+    @Input
     String copyright
 
     /**
      * The file encoding of the <code>NOTICE</code> file.
      */
+    @Optional
+    @Input
     String encoding
 
     private static final String NOTICE_PATH = "META-INF/NOTICE"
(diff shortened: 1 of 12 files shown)


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheNoticeResourceTransformer.groovy
```diff
@@ -34,34 +36,48 @@ import java.text.SimpleDateFormat
  * @author John Engelman
  */
 class ApacheNoticeResourceTransformer implements Transformer {
+
+    @Input
     Set<String> entries = new LinkedHashSet<String>()
```
wolfs, line 41:
```suggestion
    private Set<String> entries = new LinkedHashSet<String>()
```

It seems like this is modified while running the transformer, so I think this should be made private.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheNoticeResourceTransformer.groovy
```diff
@@ -34,34 +36,48 @@ import java.text.SimpleDateFormat
...
 
+    @Input
     Map<String, Set<String>> organizationEntries = new LinkedHashMap<String, Set<String>>()
```
wolfs, line 44:
```suggestion
    private Map<String, Set<String>> organizationEntries = new LinkedHashMap<String, Set<String>>()
```

It seems like this is modified while running the transformer, so I think this should be made private.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/IncludeResourceTransformer.groovy
```diff
@@ -33,8 +37,12 @@ import org.codehaus.plexus.util.IOUtil
  * @author John Engelman
  */
 public class IncludeResourceTransformer implements Transformer {
+
+    @InputFile
+    @PathSensitive(PathSensitivity.NAME_ONLY)
```
wolfs, line 42:
```suggestion
    @PathSensitive(PathSensitivity.NONE)
```

I think the name doesn't matter here.

wolfs (commented):
I left some comments.

I think this looks good and fixes the deprecation warnings with Gradle 7.0 in the production code.

helfper:
@wolfs Thanks for reviewing the PR. I've applied your suggestions, but I'm fine if #646 gets merged instead of this. I like how you added the deprecation checks globally.

PR #655, opened 2021-03-29
Use protected superclass fields via super because of GROOVY-9292

maxm123:
The use of the abstract fields in the ServiceFileTransformer without super results in an error in newer Groovy versions since [PR 1050](https://github.com/apache/groovy/pull/1050). For the current Gradle 7.0 RC, that looks like this:
```
Caused by: groovy.lang.MissingPropertyException: No such property: count for class: com.github.jengelman.gradle.plugins.shadow.transformers.ServiceFileTransformer
        at com.github.jengelman.gradle.plugins.shadow.transformers.ServiceFileTransformer$ServiceStream.append(ServiceFileTransformer.groovy:110)
        at ...
```
Apart from that, the plugin is usable with Gradle 7.0, as far as I can tell.


breskeby:
Any concrete plans for the 6.2 release?

johnrengelman:
Working on it now Rene.

breskeby:
Nice! Thanks John!

PR #657, opened 2021-04-07
Remove imports of old maven plugin classes for Gradle 7.0 build compat

### LINKED ISSUE #658: Incompatibility with JDK 16

### Shadow Version
6.1.0

### Gradle Version
7.0-rc-2

### Expected Behavior

The task `shadowjar` working when using JDK 16 , works when using JDK 15
### Actual Behavior

It failes
-----------------------------------------------------------------------------------------------
The stacktrace: https://gist.github.com/tofikarz/0b3ae6e45316bbc825acb3d175677ed1

maxm123:
The 'maven' plugin which is deprecated for quite some time is finally [removed](https://github.com/gradle/gradle/commit/53d7c1bcf272ff9bf1fc8e2145b458cb9072d06b) in Gradle 7.0. This also [affects](https://github.com/gradle/gradle/commit/eafc8d71fc2cf1bcfe565432d686cc25c06aedb9) the related classes (their replacements already live under org.gradle.api.publish.maven).

The ShadowJavaPlugin class currently imports them which makes it incompatible with Gradle 7 at compile time. By removing the imports and using Groovy's def variable declaration, the search for those old classes is deferred to runtime, only being called when the 'maven' plugin is applied and the classes are there as expected. So this change is compatible with use of the plugin in older Gradle versions.

As Gradle emits warnings for the jcenter repository, claiming it to be removed in Gradle 8, I followed their suggestion to change it to mavenCentral.

With this change, I am able to build and use this plugin with Gradle 7.
Fixes #658

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,10 +2,7 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocatio
 
 buildscript {
     repositories {
-        jcenter()
-        maven {
-            url "https://plugins.gradle.org/m2/"
-        }
+        gradlePluginPortal()
     }
     dependencies {
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.1'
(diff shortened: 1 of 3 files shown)


krolik-exe:
You Can just remove the jcenter() without replacement as GradlePluginPortal() has everything , see #659

maxm123:
Thanks for pointing out, @krolik-exe!

PR #661, opened 2021-04-09
Allow for opt-out of the java shadowRuntimeElements variant

### LINKED ISSUE #651: Unable to publish only unshadowed artifacts
I've used 5.2.0 before and then decided to migrate to 6.1.0
What I found is that `project.components.java` is updated with shadowJar configurations here:
https://github.com/johnrengelman/shadow/blame/b3677a26c706aec9aa7b035b9d36a091523aff97/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy#L56

So, I can't use maven-publish to publish only raw, unshadowed JARs: publication JAR conflict happens.
Please, make adding shadowJar things to java software component optional: I'd enable this option in project config.
Thank you!

m8nmueller:
Add a extension property to regulate whether a new variant is added to the java component. I must admit I don't know exactly why the variant was added. The [commit message](https://github.com/johnrengelman/shadow/commit/5c572a68bebf712a99854685d6c2834d146f9924) was only about support for Gradle 5 -> 6. So this is just a proposal.
Closes #651

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
@@ -12,10 +12,12 @@ class ShadowExtension {
 
     CopySpec applicationDistribution
     Project project
+    boolean addJavaVariants
 
     ShadowExtension(Project project) {
         this.project = project
         applicationDistribution = project.copySpec {}
+        addJavaVariants = true
     }
 
     void component(MavenPublication publication) {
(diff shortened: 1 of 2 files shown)


maxm123:
Hi @johnrengelman,
I was trying to look after another "problem" if you can call it one: The shadowRuntimeElements configuration lists the shadow jar artifact twice (see native outgoingVariants task) - one inherited from the shadow configuration, the other one being the output. Looking at the code, I wondered why shadowRuntimeElements is needed at all. Wouldn't it be possible to just reconfigure the shadow configuration accordingly?

Goooler:
Handled by #1662.

PR #708, opened 2021-09-08
Fix for shadowJar out-of-date with configuration caching

mathjeff:


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java
```diff
@@ -31,7 +31,7 @@
 
     private List<Transformer> transformers;
     private List<Relocator> relocators;
-    private transient List<Configuration> configurations;
+    private List<FileCollection> configurations;
```
mathjeff, line 34: I'm not sure whether the removal of `transient` can affect correctness of outputs; I was surprised that I didn't notice any differences in output between my first and second builds even though the second build was not UP-TO-DATE.

However, removing `transient` does seem to allow the task to be up-to-date in the second build.

cc @eskatos since it seems to have been added here: https://github.com/johnrengelman/shadow/commit/c9e30f10aecd69600213a2853296d9e6175a715d#diff-d5abeaae583093a8a92ef53ffe7f5a21cf185cf0d5e3a76edd236fa26dbf7e9fR34

eskatos: Good call! Looks good to me

eskatos (approved):

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java
mathjeff: Thanks!

mathjeff:
It looks like I don't have permission to submit this or run checks against it; if anybody that does have permission would like to take a look that would be super cool

mathjeff:
Thanks!

PR #724, opened 2021-11-12
Mark the Log4j2PluginsCacheFileTransformer as cacheable.

staktrace:
This transfomer doesn't take any custom inputs; it has
deterministic behaviour based on the files getting shaded.
This means it doesn't pose any additional requirements on the
cacheability of the overall ShadowJar task. Therefore we
should be able to mark it cacheable.


staktrace:
I would also *love* to see a new release including this change sooner rather than later, as it would provide a rather large speedup for us. Thank you for all your work on this plugin, I really appreciate it!

PR #756, opened 2022-02-15
Add the ability to rename annotations added by the Scala compiler.

shawjef3:
This is for https://github.com/johnrengelman/shadow/issues/146. I haven't tested the new functionality, but existing tests pass.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
@@ -1,5 +1,6 @@
 package com.github.jengelman.gradle.plugins.shadow.tasks
 
+import com.eed3si9n.jarjarabrams.scalasig.ScalaSigClassVisitor
 import com.github.jengelman.gradle.plugins.shadow.ShadowStats
 import com.github.jengelman.gradle.plugins.shadow.impl.RelocatorRemapper
 import com.github.jengelman.gradle.plugins.shadow.internal.GradleVersionUtil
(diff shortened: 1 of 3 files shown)


shawjef3:
@johnrengelman The build failed due to a temporary godaddy failure. Please retrigger the build.

johnrengelman:
Can you merge master to this. There’s a fix there.

shawjef3:
@johnrengelman it passed. Thanks!

shawjef3:
@johnrengelman I'm wondering if you've run into this. I want to use this patch with a project I have locally. I publish it.

`./gradlew :publishPluginPublicationToMavenLocal`

In my project I use `pluginManagement` in `settings.gradle` to have Gradle use my local published snapshot instead of the official Shadow publication. Then I run `:shadowJar` on my project. I get,

```
> Failed to apply plugin [class 'com.github.jengelman.gradle.plugins.shadow.ShadowJavaPlugin']
   > Could not create plugin of type 'ShadowJavaPlugin'.
      > The constructor for type ShadowJavaPlugin should be annotated with @Inject.
```

I looked at the class in the jar in my local repo, and it does have the annotation. However, when I debug Gradle's `InjectUtil`, the annotation is missing from the class. It's like Gradle's class loader is stripping it.

shawjef3:
I was able to use `((GeneratedClassImpl) type).generatedClass.getConstructors()[0].getAnnotations()` to find that the actual annotation on the constructor is `@shadow.javax.inject.Inject()`. So it seems the publication is shadowing annotations.

shawjef3:
I figured out what's happening. Something about my patch makes annotation classes be shaded also. This could be desirable but in this case it's not. I'll figure out how to make it only shade that Scala annotations.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.groovy
```diff
@@ -348,7 +349,7 @@ class ShadowCopyAction implements CopyAction {
             // that use the constant pool to determine the dependencies of a class.
             ClassWriter cw = new ClassWriter(0)
 
-            ClassVisitor cv = new ClassRemapper(cw, remapper)
+            ClassVisitor cv = new ScalaSigClassVisitor(new ClassRemapper(cw, remapper), remapper.scalaRenamer)
```
Goooler, line 352: We should compat Kotlin as well, need further work like #1539.

PR #765, opened 2022-04-10
Fix value not set when rawString is true.

### LINKED ISSUE #764: Throw required value not set exception when rawString is set to true

### Shadow Version
7.1.2
### Gradle Version
7.4.2
### Expected Behavior
work well when rawString is true.
### Actual Behavior
Throw required value not set exception when rawString is set to true.
### Gradle Build Script(s)
```kotlin
tasks.withType<ShadowJar> {
    val prefix = "shaded"
    listOf(
        "io.netty",
    ).forEach { relocate(it, "$prefix.$it") }
    relocate(
        SimpleRelocator(
            "META-INF/native/libnetty_transport_native_epoll_x86_64.so",
            "META-INF/native/lib${prefix.replace(".", "_")}_netty_transport_native_epoll_x86_64.so",
            emptyList(),
            emptyList(),
            true
        )
    )
}
```

qian0817:
fix #764.
If rawString is true, pattern and shadedPattern will be null, gradle will throw exception if not have `@Optional` annotation.


qian0817:
@johnrengelman can you take a look？

PR #769, opened 2022-05-17
Do not add a dependencies block if it's already there

desiderantes:
Sometimes the publication already has a pom, so adding a node will result in a duplicated dependencies block

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
@@ -28,8 +28,8 @@ class ShadowExtension {
 
         publication.pom { MavenPom pom ->
             pom.withXml { xml ->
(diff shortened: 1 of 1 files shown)


ind1go:
This is just what I was looking for! Do snapshots of the plugin get published that I can use pre-release?

simPod:
This broke CI https://app.circleci.com/pipelines/github/johnrengelman/shadow/224/workflows/e25665d9-0a3f-49a2-ba6d-221b771d890f/jobs/713

simPod:
Fixed in https://github.com/johnrengelman/shadow/pull/791

netomi:
> Fixed in #791

This PR just reverts the behavior as it was prior to the conflicting commit.

simPod:
Exactly. But since this PR broke the lib, it is better to revert rather then leave it as is.

netomi:
PR #798 fixes the issue in a correct way by avoiding duplicate dependencies nodes.

PR #793, opened 2022-09-16
Fix RelocatorRemapper: do not map inner class name if not changed

### LINKED ISSUE #792: Relocation breaks inner classes which contains '$' in name
The following Kotlin code is compiled with an inner class named `$serializer` by the Kotlin compiler. 
```kotlin
package foo
@Serializable // This annotation instructs the compiler to generate an extra inner class named `$serializer`
enum class MyEnum {
    VALUE1
}
```
The qualified name of the inner class will be `foo.MyEnum$$serializer`. 

If relocation is enabled, shadow runs `RelocatorRemapper` on the class. In `org.objectweb.asm.commons.Remapper`, it maps inner class names by substring-ing the name after the last `$`. So the resultant qualified name for the example will be `foo.MyEnum$serializer` (Note one `$` is missing). This breaks Kotlin's class metadata and will cause `AssertionError` in runtime. See relevant issues like mamoe/mirai#2230.

----

Here's the `org.objectweb.asm.commons.Remapper.mapInnerClassName` for your reference.
```java
  /**
   * Maps an inner class name to its new name. The default implementation of this method provides a
   * strategy that will work for inner classes produced by Java, but not necessarily other
   * languages. Subclasses can override.
   *
   * @param name the fully-qualified internal name of the inner class.
   * @param ownerName the internal name of the owner class of the inner class.
   * @param innerName the internal name of the inner class.
   * @return the new inner name of the inner class.
   */
  public String mapInnerClassName(
      final String name, final String ownerName, final String innerName) {
    final String remappedInnerName = this.mapType(name);
    if (remappedInnerName.contains("$")) {
      int index = remappedInnerName.lastIndexOf('$') + 1;
      while (index < remappedInnerName.length()
          && Character.isDigit(remappedInnerName.charAt(index))) {
        index++;
      }
      return remappedInnerName.substring(index);
    } else {
      return innerName;
    }
  }
```

### Shadow Version

7.1.2

### Gradle Version

7.3.0

### Expected Behavior

Code runs successfully.

Him188:
Fixes #792

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/impl/RelocatorRemapper.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/impl/RelocatorRemapper.groovy
@@ -115,4 +115,32 @@ class RelocatorRemapper extends Remapper {
         mapPath(path.pathString)
     }
 
+    @Override
+    String mapInnerClassName(String name, String ownerName, String innerName) {
+        final String remappedInnerName = this.mapType(name)
+
+        if (remappedInnerName == name) {
+            return innerName
+        } else {
+            int originSplit = name.lastIndexOf('/')
+            int remappedSplit = remappedInnerName.lastIndexOf('/')
+            if (originSplit != -1 && remappedSplit != -1) {
+                if (name.substring(originSplit) == remappedInnerName.substring(remappedSplit)) {
+                    // class name not changed
+                    return innerName
+                }
+            }
+        }
+
+        if (remappedInnerName.contains('$')) {
+            int index = remappedInnerName.lastIndexOf('$') + 1
+            while (index < remappedInnerName.length()
+                    && Character.isDigit(remappedInnerName.charAt(index))) {
+                index++
+            }
+            return remappedInnerName.substring(index)
+        } else {
+            return innerName
+        }
+    }
 }


Him188:
Note that [the relevant PR for asm](https://gitlab.ow2.org/asm/asm/-/merge_requests/363) has been merged. When a new version of asm with this commit is published, we can simply upgrade to the new version.

ArcticLampyrid:
not necessary to merge

we've upgradled to asm v9.4, and this is fixed by upstream.

> 2 October 2022: ASM 9.4 (tag ASM_9_4)
> - new Opcodes.V20 constant for Java 20
> - more checks in CheckClassAdapter
> - Javadoc improvements and fixes
> - module-info classes can be built without Gradle and Bnd
> - parent POM updated to org.ow2:ow2:1.5.1
> - bug fixes
>   * 317977: CheckClassAdapter is no longer transparent for MAXLOCALS
>   * 317981: Add public getDelegate method to all visitor classes
>   * Analyzer does not compute optimal maxLocals for static methods
>   * Fix SignatureWriter when a generic type has a depth over 30
>   * ***Skip remap inner class name if not changed in Remapper***

johnrengelman:
Reverting this in 8.1.1. 
Thanks @ArcticLampyrid

PR #773, opened 2022-06-10
Collision logging, Transformers for JSON and Standard Files

Master-Code-Programmer:
_Last edit: 2023-11-01_

Hi, I fitted the collsion_logging merge request from chapmajs to your current source code (https://github.com/johnrengelman/shadow/pull/126).
Also I followed your wish to use typed parameters (which was really needed, because after adding them, some problems became obvious).

After that I used the modified plugin and with the warnings I could fix my problems instantly. Great!

The only problem was, that I was flooded with warnings about colliding META-INF/NOTICE, META-INF/license.txt, readme.txt and similar files. So I added a transformer, StandardFilesMergeTransformer, which takes these standard files, which exist in all big projects in big numbers, and merges them.
Since these are all pretty primitive files regarding their structure, no special YAML or JSON files, I just concatenate their contents the following way:

* Standard files are "META-INF/license", "META-INF/notice",  "META-INF/readme" or "readme". Which may have the following extensions: none, "txt", "md", "htm" or "html".
* If a standard file is found (case insensitive), its content is read, stripped of leading and trailing newlines, and then searched in the list of already existing contents of that file. The search is case insensitive. 
   * If no existing entry was found a new entry is created and its content and origin is recorded. 
   * If an existing entry is found, its list of origins is extended with the origin of the content.
* Later, the transformed entries are written like that:
```
    Origins: attoparser-2.0.5.RELEASE.jar, commons-codec-1.15.jar

                                 Apache License
                           Version 2.0, January 2004
    ...
    ================================================================================
    Origins: flyway-core-8.5.4.jar, flyway-mysql-8.5.4.jar

    Copyright (C) Red Gate Software Ltd 2010-2022

    Licensed under the Apache License, Version 2.0 (the "License");
    ...
```
* The StandardFilesMergeTransformer transformer is added as default. It can be removed with "removeDefaultTransformers".
   I added it as default because otherwise:
   * In projects with many dependencies the user gets flooded with information about duplicate entries
            like "META-INF/notice.txt", "META-INF/license.txt"...
   * Important licensing information written in META-INF/license.txt and other files may be lost.
   * Helpful information written in readme files may be lost.
   * The merging of this plain text files is safe, there is no important logic to follow. Merged HTML may not look that good, but it works.

Edit:
[...]

PR #796, opened 2022-10-11
Add version of UnusedTracker that supports full shrinking using R8.

netomi:
Updated PR for #566 
Adds support for #565

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultR8Configuration.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultR8Configuration.groovy
@@ -1,0 +1,16 @@
+package com.github.jengelman.gradle.plugins.shadow.internal
+
+class DefaultR8Configuration implements R8Configuration {
+
+    private final Collection<String> additionalRules = new ArrayList<>();
+
+    @Override
+    Collection<String> getRules() {
+        return additionalRules
+    }
+
+    @Override
+    void rule(String rule) {
+        additionalRules.add(rule)
+    }
+}
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultR8Configuration.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultR8Configuration.groovy
@@ -2,7 +2,8 @@ package com.github.jengelman.gradle.plugins.shadow.internal
 
 class DefaultR8Configuration implements R8Configuration {
 
-    private final Collection<String> additionalRules = new ArrayList<>();
+    private final Collection<String> additionalRules    = new ArrayList<>()
+    private final Collection<File>   configurationFiles = new ArrayList<>()
 
     @Override
     Collection<String> getRules() {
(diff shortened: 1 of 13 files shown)


netomi:
Changes that I have performed since the last PR:

* Extract UnusedTracker as an abstract base class for different implementations
* Existing code is moved to UnusedTrackerWithJDependency and used by default
* add an useR8 option to the ShadowSpec to enable R8 minimization
* use the latest R8 release available on the google maven repo
* separate unit tests in MinimizeR8Spec
* Implement UnusedTrackerWithR8 in java instead of groovy to avoid groovy madness in combination with R8

I applied all comments apart from the dependency on R8, its still implementation. I wanted to keep it like that to be able to easily write unit tests using the new shrinking mode. However I am perfectly fine to change it to a compileOnly dependency and check if R8 is available at runtime and use it in that case.

JD-CSTx:
@netomi Yeah: Your CI tests fail for the same reason as in my pull request: #773. Link to comment: https://github.com/johnrengelman/shadow/pull/773#issuecomment-1268201381.

netomi:
> @netomi Yeah: Your CI tests fail for the same reason as in my pull request: #773. Link to comment: [#773 (comment)](https://github.com/johnrengelman/shadow/pull/773#issuecomment-1268201381).

I have seen that, it comes from this commit:

https://github.com/johnrengelman/shadow/commit/38566d50345de45d25b1404669f9e1d6856e5e99

Seems easy to fix, will create a PR.

JD-CSTx:
@netomi I'm not so sure. The original PR author had a valid case for his/her PR: "Do not add a dependencies block if it's already there". But the PR seemed to be incomplete. I saw someone else just reverting it, I don't think that's a good idea, because the idea behind it is valid.

netomi:
> @netomi I'm not so sure. The original PR author had a valid case for his/her PR: "Do not add a dependencies block if it's already there". But the PR seemed to be incomplete. I saw someone else just reverting it, I don't think that's a good idea, because the idea behind it is valid.

the change assumes that if a node 'dependencies' exists, it is never empty as it accesses the item at index 0. In the tests you can see that the node is actually empty and thus fails with an index out of bounds exception.

netomi:
The build issues wrt publication is fixed in PR #798

JD-CSTx:
> The build issues wrt publication is fixed in PR #798

Big thanks.

netomi:
Updated the documentation and added possibility to specify additional rules via the shadow spec. I would consider this PR to be finished from my side unless there is some suggestion for change. With the additional rules that can be specified via the shadow spec, this method should allow enough flexibility to handle more complex scenarios.

netomi:
A more realistic use case of the minimization with R8 can be tested here: https://github.com/netomi/uom/tree/shadow-dependencies

`./gradlew clean shadowJar`

In this project 2 classes from external dependencies are used (from guava and hibernate). Without minimization or the previous one, a huge number of classes is copied in the shadow jar. With R8 enabled, the copied dependencies is minimal as expected.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultR8Configuration.groovy
```diff
@@ -0,0 +1,16 @@
...
+
+    private final Collection<String> additionalRules = new ArrayList<>();
+
+    @Override
+    Collection<String> getRules() {
+        return additionalRules
+    }
+
+    @Override
+    void rule(String rule) {
```
gildor, line 13: It would be nice to allow load rules from config file, maybe something like `rules(file("proguard.conf"))`

netomi: Makes sense, will add this.

PR #797, opened 2022-10-11
Fix retrieval of dependencies node after #769

netomi:
The PR changes the logic to always retrieve the first child of the dependencies node it it was present.
This fails sometimes ofc if the list of dependencies is empty.
Furthermore, the PR was adding dependencies to the first encountered dependency instead of the dependency list.

This PR fixes it.


netomi:
PR has wrong base, closing it and recreate based on master

PR #824, opened 2023-02-25
Support config cache

Goooler:
- Closes #687.
- Closes #763.
- Closes #775.

This PR depends on #823.

Issue #775, opened 2022-06-23
Task `installShadowDist` does not work with Gradle Configuration Cache enabled


jvmusin:
### Shadow Version

7.1.2

### Gradle Version

7.4.2

### Expected Behavior

`runShadow` task runs with the gradle configuration cache enabled

### Actual Behavior

Gradle produces an error saying the configuration cache requirements are not met by the task `installShadowDist`

> - Task `:installShadowDist` of type `org.gradle.api.tasks.Sync`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.

### Gradle Build Script(s)

```
import org.jetbrains.kotlin.gradle.tasks.KotlinCompile

plugins {
    application
    kotlin("jvm") version "1.7.0"
    id("com.github.johnrengelman.shadow") version "7.1.2"
}

application.mainClass.set("MainKt")
repositories.mavenCentral()

tasks.withType<KotlinCompile> {
    kotlinOptions.jvmTarget = "11"
}
```

Sample project is here:
[untitled.zip](https://github.com/johnrengelman/shadow/files/8966960/untitled.zip)

jvmusin:
Class `JavaJarExec`  with the task `runShadow` are also not compatible:

```
- Task `:samples:ktor-fatjar-sample:installShadowDist` of type `org.gradle.api.tasks.Sync`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.4.2/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:samples:ktor-fatjar-sample:runShadow` of type `com.github.jengelman.gradle.plugins.shadow.internal.JavaJarExec`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.4.2/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
[...]

Issue #763, opened 2022-04-01
AppendingTransformer blocks use of configuration cache (field `data` cannot be serialized)


autonomousapps:
### Shadow Version
7.1.2

### Gradle Version
7.4

### Expected Behavior
The configuration cache is supported. Sample invocation that fails:
```
./gradlew foo:shadowJar -m --configuration-cache
```

### Actual Behavior
Attempting to use the configuration cache fails with the following:
```
// console output
$ ./gradlew bundle-aar:shadowJar -m --configuration-cache
3 problems were found storing the configuration cache, 1 of which seems unique.
- Task `:bundle-aar:shadowJar` of type `com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar`: cannot serialize object of type 'java.io.ByteArrayOutputStream', a subtype of 'java.io.OutputStream', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.4/userguide/configuration_cache.html#config_cache:requirements:disallowed_types

// and the configuration cache error report:
cannot serialize object of type java.io.ByteArrayOutputStream a subtype of java.io.OutputStream, as these are not supported with the configuration cache (https://docs.gradle.org/7.4/userguide/configuration_cache.html#config_cache:requirements:disallowed_types)
  field data of com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer
    bean of type com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer
      fieldtransformers of com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
        task:bundle-aar:shadowJar of type com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
```

Note that this is only a _dry run_. It fails in the configuration phase, before it gets to the task action.

### Gradle Build Script(s)
```groovy
configurations {
  androidTools {
    // sdklib brings in a lot of unused code paths that can cause trouble with classpath conflicts.
    // Excluding for sanity.
    exclude group: 'com.android.tools', module: 'repository'
    exclude group: 'com.android.tools.analytics-library'
  }
}

// bundle and shade androidTools to avoid conflicting with AGP versions.
shadowJar {
  configurations = [project.configurations.androidTools]
  relocate 'com.android', 'bundle.aar.com.android'
  append 'META-INF/LICENSE.txt'
  append 'META-INF/LICENSE.md'
  append 'META-INF/NOTICE.md'
  mergeServiceFiles()
  minimize()
[...]

Issue #834, opened 2023-02-27
ConfigureShadowRelocation task is not resolvable in 8.1.0


iProdigy:
### Shadow Version
8.1.0

### Gradle Version
8.0.1

### Expected Behavior
`com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocation` should be resolvable (or clearly documented that a change has occurred)

### Actual Behavior
`Unresolved reference: ConfigureShadowRelocation`

### Gradle Build Script(s)
https://github.com/twitch4j/twitch4j/blob/master/build.gradle.kts#L163 (we will be changing to the new `enableRelocation` syntax)

Caused by https://github.com/johnrengelman/shadow/pull/824/files#diff-6b2b213db0ba7e55e2818730dd5342e67e6c957c72b54acdba3a3b52885db9e2

Seems like this should've been clearly communicated in <https://github.com/johnrengelman/shadow/releases/tag/8.1.0> if intentional breaking change

Goooler:
Use shadowJar instead, like:

```groovy
tasks.named("shadowJar", ShadowJar) {
    enableRelocation true
    relocationPrefix "shadow"
}
```
see 04b3c3c10bb1719e79fbddea636e258a53163cba.

iProdigy:
Also it seems this documentation text needs to be updated:
```
To configure automatic dependency relocation, declare a task of type ConfigureShadowRelocation and configure the target parameter to be the ShadowJar task you wish to auto configure. You will also need to declare a task dependency so the tasks execute in the correct order.
```
https://github.com/johnrengelman/shadow/blob/master/src/docs/configuration/relocation/README.md#automatically-relocating-dependencies

Proximyst:
@Goooler How are we expected to filter what dependencies are to be relocated now?

A common case might be including sub-projects which don't need relocation alongside dependencies that do. It's rather inconvenient if we need to have to include tests to ensure all packages have relocations specified before distribution.

Goooler:
`enableRelocation` is disabled by default, see

https://github.com/johnrengelman/shadow/blob/49584a72811be917b2855ab365577517aa3a18e0/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.java#L34

Proximyst:
I meant what option exists for this option now?

[...]

PR #835, opened 2023-02-28
Fix more CC compatibility issues

Goooler:
I missed something in #763 & #775, cause `--configuration-cache` flag hadn't been applied for all plugin tests, now fixed all of them.

Follow up #824.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
@@ -94,7 +94,6 @@ class ShadowApplicationPlugin implements Plugin<Project> {
 
     protected void configureInstallTask(Project project) {
         project.tasks.named(SHADOW_INSTALL_TASK_NAME, Sync).configure { task ->
-            notCompatibleWithConfigurationCache('https://github.com/johnrengelman/shadow/issues/775')
             task.doFirst {
                 if (task.destinationDir.directory) {
                     if (task.destinationDir.listFiles().size() != 0 && (!new File(task.destinationDir, 'lib').directory || !new File(task.destinationDir, 'bin').directory)) {
@@ -106,7 +105,11 @@ class ShadowApplicationPlugin implements Plugin<Project> {
                 }
             }
             task.doLast {
-                project.ant.chmod(file: "${task.destinationDir.absolutePath}/bin/${javaApplication.applicationName}", perm: 'ugo+x')
+                task.eachFile {
+                    if (it.path == "bin/${javaApplication.applicationName}") {
+                        it.mode = 0x755
+                    }
+                }
             }
         }
     }
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
@@ -93,8 +93,9 @@ class ShadowApplicationPlugin implements Plugin<Project> {
     }
 
     protected void configureInstallTask(Project project) {
-        project.tasks.named(SHADOW_INSTALL_TASK_NAME).configure { installTask ->
-            installTask.doFirst { Sync task ->
+        project.tasks.named(SHADOW_INSTALL_TASK_NAME, Sync).configure { task ->
+            notCompatibleWithConfigurationCache('https://github.com/johnrengelman/shadow/issues/775')
+            task.doFirst {
                 if (task.destinationDir.directory) {
                     if (task.destinationDir.listFiles().size() != 0 && (!new File(task.destinationDir, 'lib').directory || !new File(task.destinationDir, 'bin').directory)) {
                         throw new GradleException("The specified installation directory '${task.destinationDir}' is neither empty nor does it contain an installation for '${javaApplication.applicationName}'.\n" +
@@ -104,7 +105,7 @@ class ShadowApplicationPlugin implements Plugin<Project> {
                     }
                 }
             }
-            installTask.doLast { Sync task ->
+            task.doLast {
                 project.ant.chmod(file: "${task.destinationDir.absolutePath}/bin/${javaApplication.applicationName}", perm: 'ugo+x')
             }
         }
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
@@ -1,33 +1,38 @@
 package com.github.jengelman.gradle.plugins.shadow
 
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import org.gradle.api.Project
 import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.artifacts.SelfResolvingDependency
+import org.gradle.api.file.RegularFile
+import org.gradle.api.provider.Provider
 import org.gradle.api.publish.maven.MavenPom
 import org.gradle.api.publish.maven.MavenPublication
-import org.gradle.util.GradleVersion
 
 class ShadowExtension {
+    private final Provider<Provider<RegularFile>> archiveFile
+    private final Provider<List<Dep>> allDependencies
 
-    void component(Project project, MavenPublication publication) {
-
-        if (GradleVersion.current() >= GradleVersion.version("6.6")) {
-            publication.artifact(project.tasks.named("shadowJar"))
-        } else {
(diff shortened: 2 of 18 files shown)


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
```diff
@@ -93,8 +93,9 @@ class ShadowApplicationPlugin implements Plugin<Project> {
     }
 
     protected void configureInstallTask(Project project) {
-        project.tasks.named(SHADOW_INSTALL_TASK_NAME).configure { installTask ->
-            installTask.doFirst { Sync task ->
+        project.tasks.named(SHADOW_INSTALL_TASK_NAME, Sync).configure { task ->
```
Goooler, line 96: Not sure this is the proper way, I just see
```java
@DisableCachingByDefault(because = "Not worth caching")
public abstract class Sync extends AbstractCopyTask 
```

src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ApplicationSpec.groovy
```diff
@@ -105,7 +105,7 @@ class ApplicationSpec extends PluginSpecification {
             runShadow {
                args 'foo'
                doFirst {
-                project.logger.lifecycle("Running application with JDK \${it.javaLauncher.get().metadata.languageVersion.asInt()}")
+                   logger.lifecycle("Running application with JDK \${it.javaLauncher.get().metadata.languageVersion.asInt()}")
```
Goooler, line 108: Use taks logger instead here.

src/test/groovy/com/github/jengelman/gradle/plugins/shadow/PublishingSpec.groovy
```diff
@@ -52,6 +52,10 @@ class PublishingSpec extends PluginSpecification {
                    }
                }
             }
+            
+            tasks.withType(org.gradle.api.publish.maven.tasks.GenerateMavenPom).configureEach {
+                notCompatibleWithConfigurationCache('https://github.com/gradle/gradle/issues/21926')
+            }
```
Goooler, line 58: Issue of `GenerateMavenPom`, we shouldn't handle it.

Goooler:
We should note these behavior changes in README or Changelog, something like
1. Replacing `ConfigureShadowRelocation` with `ShadowJar`, link #834.
2. If you are using custom `Transformer`s, disable CC in `shadowJar`.
3. If you are using `ShadowExtension.component`, disable CC for `org.gradle.api.publish.maven.tasks.GenerateMavenPom`.

src/test/groovy/com/github/jengelman/gradle/plugins/shadow/PublishingSpec.groovy
Goooler: Filed at https://github.com/gradle/gradle/issues/24097.

Goooler: fixed in 6bbc4010a0cd6e50a50a40f1c819fa6f3e8735d3

Goooler:
@eskatos Can you take a look?

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.groovy
eskatos: If the only reason for this `Sync` task to me marked as not compatible with the configuration cache is the usage of `project.ant.chmod` in its `doLast {}`  action, then this can be replaced by configuring the file mode using:

```groovy
task.eachFile {
    if (it.path == "bin/${javaApplication.applicationName}") {
        mode = 0x755
    }
}
```

Goooler: Seems can't, see https://github.com/Goooler/shadow/commit/2a91f58204ca49b720bdf2556026f71deca49b79

eskatos: The `eachFile {}` configuration should not be in the `doLast {}` action but in the task configuration.

Goooler: Same result in https://github.com/Goooler/shadow/commit/c7e3953a219a5b198f51f8563b74fe09d74c0aef

eskatos: Then something else must hold a reference to the project.
Probably `javaApplication`?

What about when you do this?
```groovy
def appName = javaApplication.applicationName
 task.eachFile {
  if (it.path == "bin/${appName}") {
    it.mode = 0x755
  }
}
```

Goooler: Tryied in https://github.com/Goooler/shadow/pull/16

eskatos: That won't work because the `AntBuilder` is not supported by CC.
You need to combine both capturing the app name in a provider and use `eachFile {}` instead of `ant.chmod`

Goooler: It works now in 3543ca6559aa81d33ed1f4d7338d150a2da3fd7c.

Goooler: Thanks!

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.groovy
```diff
@@ -3,42 +3,50 @@ package com.github.jengelman.gradle.plugins.shadow
...
             publication.artifact(project.tasks.shadowJar)
         }
 
+        final def allDependencies = project.provider {
+            project.configurations.shadow.allDependencies.collect {
+                if ((it instanceof ProjectDependency) || ! (it instanceof SelfResolvingDependency)) {
+                    new Dep(it.group, it.name, it.version)
+                }
+            }
+        }
```
Goooler, line 26: Can we move `allDependencies` into `ShadowExtension`'s constructor?

eskatos: I would say yes if you don't put the `project` reference into a field.

Another trick you could use is to keep the `Project` field but as a `transient`.
It would then be available during the configuration phase but not anymore during execution.
This could allow to remove the need to have the `Project` parameter in the `component()` function.

Goooler: I tried it in https://github.com/Goooler/shadow/pull/17, but something is missing.

eskatos: It seems Groovy closure capturing is confused by the field instead of a local variable :/
It tries to find an xml element named `allDependencies` in the document :shrug: 
This looks like a Groovy problem to me.
Maybe declaring a local variable with the field assigned to it before the `withXml` closure would help :thinking:

Goooler: https://scans.gradle.com/s/iha6smzemxmuq/tests/:test/com.github.jengelman.gradle.plugins.shadow.PublishingSpec/publish%20shadow%20jar%20with%20maven-publish%20plugin?top-execution=1
or
https://scans.gradle.com/s/glxqqw5mj2rlm/tests/:test/com.github.jengelman.gradle.plugins.shadow.PublishingSpec/publish%20shadow%20jar%20with%20maven-publish%20plugin?top-execution=1

I was stuck there for days, I'm guessing there is something still not be serialized.

Goooler: Fixed by a86645c0dcb82fcd4663c313ed5345481478a85b.

Goooler:
@johnrengelman Can we merge this and publish a new version?

johnrengelman:
I should be able to get to it tomorrow. 
thanks for all the contributions!

Goooler:
@johnrengelman Any updates?

johnrengelman:
Releasing this today @Goooler. Thanks!
Also, i've switch the repo to now use `main`. Make sure to updated your tracking branches.

Issue #856, opened 2023-04-13
PropertiesFileTransformer breaks Reproducible builds in 8.1.1


agascon:
### Shadow Version
8.1.1

### Gradle Version
Not relevant.

### Expected Behavior
PropertiesFileTransformer supports having **reproducible builds** by not introducing any timestamp dependant of the build time.

### Actual Behavior
Fix for supporting reproducible builds introduced in https://github.com/johnrengelman/shadow/commit/e46bd204917bc004c1a92c7d1ffa36250b192904 in 8.0.0 were overridden by  https://github.com/johnrengelman/shadow/commit/9bffd1e3762ebd8c149b9fd3e7cf3837c7528167 in 8.1.1

PropertiesFileTransformer is back to use Properties class and it's introducing timestamps into transformed files.

johnrengelman:
@agascon do you have something to demonstrates the issue? It doesn't appear that any of the changes were actually overridden in the merge (https://github.com/johnrengelman/shadow/blob/main/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.groovy#L185-L203). It was just 2 changes that were in the same place, but the 3-way merge accounted for it. Additionally, the tests added in https://github.com/johnrengelman/shadow/commit/e46bd204917bc004c1a92c7d1ffa36250b192904https://github.com/johnrengelman/shadow/commit/e46bd204917bc004c1a92c7d1ffa36250b192904 are still passing.

agascon:
Unfortunately the project where I detected this is private but will try to provide something.

I understand problem is in https://github.com/johnrengelman/shadow/blob/main/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.groovy#L253.  That _store_ method is not the one overridden in CleanProperties.

agascon:
build.gradle
```
import com.github.jengelman.gradle.plugins.shadow.transformers.PropertiesFileTransformer

plugins {
	id 'java'
	id 'org.springframework.boot' version '2.7.10'
	id 'io.spring.dependency-management' version '1.0.15.RELEASE'
	id 'com.github.johnrengelman.shadow' version '8.1.1'
}

group = 'com.example'
version = '0.0.1-SNAPSHOT'
sourceCompatibility = '17'

repositories {
	mavenCentral()
}

dependencies {
	implementation 'org.springframework.boot:spring-boot-starter'
[...]

PR #858, opened 2023-04-18
Fix for PropertiesFileTransformer breaks Reproducible builds in 8.1.1 #856

agascon:
Please have a look, added a test (which should fail with 8.1.1) and a fix on CleanProperties.

Removing the timestamp which default Properties class is adding could be tricky.  I changed the previous mechanism to be bit more flexible (was not working either as some blank comment line is added now), but either way don't know if is a good approach.  Please check


agascon:
Hey @johnrengelman, any chance to have a looks on this?

Goooler:
Addressed to https://github.com/Goooler/shadow/pull/102, this could be closed for now.

Goooler:
Closing due to #876.

PR #912, opened 2024-03-13
Fix invalid example in docs

staktrace:
The example as provided doesn't seem to work. Updating to what I believe is intended here.

### DIFF
--- a/src/docs/configuration/dependencies/README.md
+++ b/src/docs/configuration/dependencies/README.md
@@ -146,9 +146,9 @@ dependencies {
 
 shadowJar {
   dependencies {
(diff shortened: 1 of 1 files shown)


staktrace:
FWIW I'm using kotlin syntax. Maybe this is groovy syntax and it works in that case? I'm not sure.

Goooler (approved):

staktrace:
@Goooler I don't have permissions to merge the PR so please feel free to do that if you think the PR is fine. 🙇

Goooler:
I don't have the permission as well :)

oleg-nenashev (approved):

PR #876, opened 2023-07-20
Support Gradle 8.9 & Java 21

Goooler:
- Adopt #854.
- Adopt #871.
- Closes #856.
- Closes #858.
- Closes #879.
- Closes #888.
- Closes #892.
- Closes #894.
- Closes #911.
- Closes #919.
- Closes #924.
- Closes #925.
- Closes #927.

--------------

As a workaround, you can try out my fork, see https://plugins.gradle.org/plugin/io.github.goooler.shadow.

PR #919, opened 2024-04-30
Fix deprecation when running on Gradle 8.8

britter:
`CopyProcessingSpec.fileMode` was recently deprecated. It is replaced by
the `filePermissions` method, that was introduced in 8.3.
For that reason this change updates the build to 8.3 to be able to call
`filePermissions`. The code then uses that method if running on 8.3 or
higher. If not, it falls back to calling `fileMode`.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.3-bin.zip
 networkTimeout=10000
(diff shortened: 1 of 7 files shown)


britter:
@johnrengelman I can fix the failing tests that complain about deprecations by replacing usages of `ApplicationConventions`. I don't see a way to make the plugin work on Gradle 8.0 ([Build Scan](https://scans.gradle.com/s/mhbzbo7aixhh4/tests/task/:test/details/com.github.jengelman.gradle.plugins.shadow.ShadowPluginSpec/Compatible%20with%20Gradle%208.0?top-execution=1)). This is because the API surface of the `FileTreeElement` has changed, and the code in this plugin implements that. I don't see a way to conditionally implement this. So my proposal is to bump the min Gradle requirement to Gradle 8.3 instead.

Goooler (dismissed):

Goooler:
This is handled by #876.

Goooler:
Closing due to #876.

PR #929, opened 2024-08-01
Prepare publish of com.gradleup.shadow

Goooler:


Goooler:
Can't obtain secrets due to forks limitations, addressed to #930.

PR #948, opened 2024-08-28
Rework excluding Gradle APIs for java-gradle-plugin

Goooler:
- Ports #459.
- Reverts #675.
- Reverts d530cff65c086e1cdd666bf3b12663949ae7ffcc.

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
@@ -75,9 +75,6 @@ class ShadowJavaPlugin implements Plugin<Project> {
             shadow.configurations = [project.configurations.findByName('runtimeClasspath') ?
                                              project.configurations.runtimeClasspath : project.configurations.runtime]
             shadow.exclude('META-INF/INDEX.LIST', 'META-INF/*.SF', 'META-INF/*.DSA', 'META-INF/*.RSA', 'module-info.class')
-            shadow.dependencies {
-                exclude(dependency(project.dependencies.gradleApi()))
-            }
         }
         project.artifacts.add(ShadowBasePlugin.CONFIGURATION_NAME, project.tasks.named(SHADOW_JAR_TASK_NAME))
     }
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
@@ -54,13 +54,16 @@ class ShadowJavaPlugin implements Plugin<Project> {
             }
         }
 
-        // Remove the gradleApi so it isn't merged into the jar file.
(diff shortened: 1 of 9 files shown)


TWiStErRob:
@Goooler as far as I can understand this will break compilation of consumer projects in the following scenario:

* Multi module project
  * Module 1: Gradle plugin with shadow
  * Module 2: java library without shadow
* Module 2 depends on Module 1
* Module 2 uses a class from Module 1
* Module 2 fails to compile if the class it uses has any Gradle classes on the public surface.

Module 1:
```kotlin
class C : Task
```

Module 2:
```kotlin
fun f(c: C) {
    println(c.project.name)
}
```

Because the `gradleApi()` is no longer `api()` the consuming Module 2 will not see the superclass and therefore won't know what `.project` is.

I don't think shadow should affect users in this way, but rather it should filter the necessary views of configurations as needed, just like publishing.

(This is theoretical, haven't tested it, but I've seen similar issues in the past.)

TWiStErRob:
https://github.com/gradle/gradle/blob/v8.12.0/platforms/extensibility/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/JavaGradlePluginPlugin.java#L178

Goooler:
Ooops, I missed it's `api`! Yeah in this case, users should manually declare `gradleApi()` in the module B, such a little thing. May have to note this point in https://github.com/GradleUp/shadow/tree/main/src/docs/plugins.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
```diff
@@ -51,6 +53,19 @@ class ShadowJavaPlugin implements Plugin<Project> {
...
+            // This is required because 'java-gradle-plugin' adds gradleApi() to the 'api' configuration.
+            // See https://github.com/gradle/gradle/blob/972c3e5c6ef990dd2190769c1ce31998a9402a79/subprojects/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/JavaGradlePluginPlugin.java#L161
+            project.configurations.named(JavaPlugin.API_CONFIGURATION_NAME) {
+                it.dependencies.remove(project.dependencies.gradleApi())
+            }
+            // Compile only gradleApi() to make sure the plugin can compile against Gradle API.
+            project.configurations.named(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME) {
+                it.dependencies.add(project.dependencies.gradleApi())
+            }
+        }
```
Goooler, line 68: I did the same thing with the logic from `com.gradle.publish.PublishTaskShadowAction`.

```java
private void manipulateDependencies(ConfigurationContainer configurations) {
    Dependency gradleApi = project.getDependencies().gradleApi();

    Configuration api = configurations.findByName("api");
    if (api != null) {
        api.getDependencies().remove(gradleApi);
    }

    Configuration compileOnly = configurations.findByName("compileOnly");
    if (compileOnly != null) {
        compileOnly.getDependencies().add(gradleApi);
    }
}
```

TWiStErRob: Ah, nice!

PR #960, opened 2024-09-08
Apply legacy plugin

### LINKED ISSUE #958: Apply an empty `com.github.johnrengelman.shadow` plugin to simplfy migration to the new plugin id `com.gradleup.shadow`

With the rename of this plugin, existing build logic that reacts to the application of the `com.github.johnrengelman.shadow` plugin will no longer be executed.

If this plugin was applying an empty plugin with this plugin id, the logic would still be executed since currently `com.gradleup.shadow` is a drop-in replacement for the previous plugin.

The main downside of doing this is that there is no clear path to deprecate and finally remove that behaviour while warning users that may rely on it without realising it as Gradle does not have APIs for this.

### Upstream changes

Gradle will be releasing a version of the `plugin-publish` plugin that accounts for this new plugin id because without it, some plugin publishing opinions are not enforced when applying `shadow`.

If there is agreement to do this, I would be willing to submit a patch for it.

ljacomet:
Closes #958 

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,6 +59,7 @@ tasks.named('shadowJar', ShadowJar) {
     from rootProject.file('LICENSE')
     from rootProject.file('NOTICE')
     enableRelocation true
+    archiveClassifier = ''
 }
 
 tasks.named('ideaModule') {
--- a/src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy
+++ b/src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy
@@ -31,7 +31,7 @@ class ShadowPluginSpec extends PluginSpecification {
 
         then:
         project.plugins.hasPlugin(ShadowPlugin)
-        project.plugins.hasPlugin("com.github.johnrengelman.shadow")
+        project.plugins.hasPlugin(LegacyShadowPlugin)
 
         and:
         assert !project.tasks.findByName('shadowJar')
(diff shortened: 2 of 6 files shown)


src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy
```diff
@@ -31,6 +31,7 @@ class ShadowPluginSpec extends PluginSpecification {
 
         then:
         project.plugins.hasPlugin(ShadowPlugin)
+        project.plugins.hasPlugin("com.github.johnrengelman.shadow")
```
Goooler, line 34:
```suggestion
        project.plugins.hasPlugin(LegacyShadowPlugin)
```

Goooler (approved):

src/test/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPluginSpec.groovy
```diff
@@ -31,6 +31,7 @@ class ShadowPluginSpec extends PluginSpecification {
 
         then:
         project.plugins.hasPlugin(ShadowPlugin)
+        project.plugins.hasPlugin(LegacyShadowPlugin)
```
Goooler, line 34:
```suggestion
        project.plugins.hasPlugin(com.github.jengelman.gradle.plugins.shadow.legacy.LegacyShadowPlugin)
```

build.gradle
```diff
@@ -59,6 +59,7 @@ tasks.named('shadowJar', ShadowJar) {
     from rootProject.file('LICENSE')
     from rootProject.file('NOTICE')
     enableRelocation true
+    archiveClassifier = ''
```
Goooler, line 62: This leads to #965.

Issue #710, opened 2021-09-13
Shadow Jar for modular project with module-info


palexdev:
### Shadow Version
Version 7.0.0

### Gradle Version
Version 7.2

### Gradle Build Script(s)
```java
jar {
    manifest {
        attributes(
                'Bundle-Name': project.name,
                'Bundle-Description': "...",
                'Bundle-SymbolicName': '...',
        )
    }
    enabled = false
    dependsOn(shadowJar)
}

shadowJar {
    mergeServiceFiles()
    archiveClassifier.set(null)
    configurations = [project.configurations.shadow]
}
```

I'm trying to build a shadow jar for my library that depends on another library, I'd like to ship my jar with the dependency already included.  
The issue is that my project is a modular project, there's the library itself that is the "main" module and then there's a preview of the library which is the "demo" module  
Both modules have their respective module-info.java, when I build the shadow jar the build fails because the "demo" module cannot find the required "main" module. In fact if I open the shadow jar there's no module info in the archive

Marcono1234:
`module-info.class` is excluded by default, see #352. Unfortunately the [documentation](https://imperceptiblethoughts.com/shadow/getting-started/#default-java-groovy-tasks) was apparently not updated to reflect that.

aoli-al:
Is it possible to include `module-info.class`? I tried `include "**/module-info.class"` but it doesn't work.

bademux:
@johnrengelman  [the decision to exclude](https://github.com/johnrengelman/shadow/issues/352#issuecomment-421812807) `module-info` was unfortunate.
The plugin is extremely useful to hackfix different legacy libs from (apache) cemetery to use modules.

Please leave `module-info` from the project as-is, but skip it from dependancies.
for now, my hackfix to restore module and expose the jar
```java
configurations {
    moduleJar {
        canBeConsumed = true
        canBeResolved = false
    }
}

task appendModueInfo(type: Jar, dependsOn: tasks.shadowJar) {
    group 'shadow'

    from zipTree(tasks.shadowJar.archiveFile)
    from tasks.compileJava.destinationDirectory.file('module-info.class')
    archiveBaseName.set(tasks.shadowJar.archiveBaseName)
    archiveVersion.set(tasks.shadowJar.archiveVersion)
    archiveClassifier.set('module')
    duplicatesStrategy = DuplicatesStrategy.FAIL
}

artifacts {
    moduleJar(tasks.appendModueInfo) {
[...]

PR #962, opened 2024-09-09
Collision logging, Transformers for JSON and Standard Files

Master-Code-Programmer:
_This is the PR #773, but on a new, rebased branch (the original PR was made 2 years ago)._

Hi, I fitted the collsion_logging merge request from chapmajs to your current source code (https://github.com/johnrengelman/shadow/pull/126).
Also I followed your wish to use typed parameters (which was really needed, because after adding them, some problems became obvious).

After that I used the modified plugin and with the warnings I could fix my problems instantly. Great!

The only problem was, that I was flooded with warnings about colliding META-INF/NOTICE, META-INF/license.txt, readme.txt and similar files. So I added a transformer, StandardFilesMergeTransformer, which takes these standard files, which exist in all big projects in big numbers, and merges them.
Since these are all pretty primitive files regarding their structure, no special YAML or JSON files, I just concatenate their contents the following way:

* Standard files are "META-INF/license", "META-INF/notice",  "META-INF/readme" or "readme". Which may have the following extensions: none, "txt", "md", "htm" or "html".
* If a standard file is found (case insensitive), its content is read, stripped of leading and trailing newlines, and then searched in the list of already existing contents of that file. The search is case insensitive. 
   * If no existing entry was found a new entry is created and its content and origin is recorded. 
   * If an existing entry is found, its list of origins is extended with the origin of the content.
* Later, the transformed entries are written like that:
```
    Origins: attoparser-2.0.5.RELEASE.jar, commons-codec-1.15.jar

                                 Apache License
                           Version 2.0, January 2004
    ...
    ================================================================================
    Origins: flyway-core-8.5.4.jar, flyway-mysql-8.5.4.jar

    Copyright (C) Red Gate Software Ltd 2010-2022

    Licensed under the Apache License, Version 2.0 (the "License");
    ...
```
* The StandardFilesMergeTransformer transformer is added as default. It can be removed with "removeDefaultTransformers".
   I added it as default because otherwise:
   * In projects with many dependencies the user gets flooded with information about duplicate entries
            like "META-INF/notice.txt", "META-INF/license.txt"...
   * Important licensing information written in META-INF/license.txt and other files may be lost.
   * Helpful information written in readme files may be lost.
   * The merging of this plain text files is safe, there is no important logic to follow. Merged HTML may not look that good, but it works.

* Added JsonTransformer for JSON files, based on [JsonTransformer](https://github.com/LogicFan/shadow-json-transformer/blob/main/plugin/src/main/groovy/com/github/logicfan/gradle/shadow/transformers/JsonTransformer.groovy) from @[LogicFan](https://github.com/LogicFan). Their code is also under [Apache License 2](https://github.com/LogicFan/shadow-json-transformer). Extended it, so it can be applied to multiple files.
* Lowered the logging level for colliding META-INF/MANIFEST.MF files, from warning to debug, so the user doesn't get flooded with unimportant messages.
* In discussion #710 it came to light that some people need the original `module-info.class` _not_ to be removed, which is also a possible source for file collision problems. While it's hard to find a nice, quick solution for that, I at least added an option to include the module-info files and list their contents (listing appears only if `module-info.class`'s aren't excluded).
The option is: `allowModuleInfos()`, and used in action in `def 'Tests the info about the module-info.class, if removed from standard excludes'()`.

**With that changes, any user should be capable to build a fat JAR with more than one **Spring (Boot)** dependency, right out of the box. Otherwise it's a real pain to find out which files collide, and how to merge colliding JSON files, since  that is the only colliding file type, which misses a merging transformer from this plugin.**
And JSON is _very_ common!


Goooler:
Sorry for the delay. Can you rebase again? I changed a lot on the main...

Master-Code-Programmer:
> Sorry for the delay. Can you rebase again? I changed a lot on the main...

Did the rebase.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
```diff
@@ -100,7 +102,18 @@ class ShadowJavaPlugin implements Plugin<Project> {
...
+             Remove excludes like this:
+             shadowJar {
+               ...
+               allowModuleInfos()
+             }
+             */
+            def excludes = ['META-INF/INDEX.LIST', 'META-INF/*.SF', 'META-INF/*.DSA', 'META-INF/*.RSA']
+            if (!shadow.isAllowModuleInfos()) {
+                excludes.add(MODULE_INFO_CLASS)
+            }
```
Goooler, line 115: I see so many changes are addressed in this PR, supporting including `module-info.class` is one of them, right? Can we extract this point as a separated PR?

.gitignore
```diff
@@ -13,7 +13,8 @@ classes/
 node_modules/
 yarn-error.log
 src/docs/.vuepress/dist/
+bin/
 .DS_Store
 jd-gui.cfg
-bin/
 .vscode/
+package-lock.json
```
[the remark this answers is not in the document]
Master-Code-Programmer: Done. Removed the change.

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy
Master-Code-Programmer: Well, the problem is: Including the `module-info.class` isn't directly a problem. While it's non-intuitive to include it, it's already possible.
The real problem comes into play if you unknowingly add a dependency which contains also a `module-info.class`. Meaning: You have a **very**, _really_ problematic file collision. Which is what this patch is about.

So I added a listing, to inform about the `module-info.class`s content. But only if the user wants to include it. Because in the default settings, the module-info is excluded anyway.
And after much exploration I gave up, and have to admit that johnrengelman was completely right in #710: The only clean way allow the module-info, is with an extra option.

But I already thought: Maybe I should only list the module-info's content if there is a file collision? On the other hand: That file is so important if included, I decided to list it always.

Master-Code-Programmer:
Converted this to a draft, I have to rewrite this in Kotlin.

Goooler:
I'm closing this for now. Feel free to open a new one for the Kotlin codebase.

Master-Code-Programmer:
@Goooler 
> Feel free to open a new one for the Kotlin codebase.

Well, I was just starting, did a bit of research, and found out that you just did already that, what was the starting point of this PR in PR #1552. You merged it on 2025-07-31, so a little over a month ago, this plugin got finally what I always needed: An alert if there are file collisions! (Tested it, works great)

Now, what I only miss are default merge-transformers for the most used file types: JSON, XML and properties files. So, standard file types are automatically merged on file collisions (all optionally of course).

Is there something in the works, and would you accept such a PR?

Goooler:
I'm wondering what's the merge processing look like. Can you show me an example for json?

Master-Code-Programmer:
Well, the merging works just like XML, where you merge it like this:
```groovy
transform(XmlAppendingTransformer) {
    resource = 'test.xml'
}
```

In JSON it's:
```groovy
transform(JsonAppendingTransformer) {
    paths = ['META-INF/spring-configuration-metadata.json',
             'META-INF/additional-spring-configuration-metadata.json']
}
```
I just have to dig a little bit deeper to find the Kotlin JSON merge transformer.

Goooler:
This looks helpful for #1489. PRs are welcome!

You can replace Gson with Moshi, it compats both Java and Kotlin.

PR #991, opened 2024-09-29
Apply legacy plugin last, and declare capabilities for old plugins

### LINKED ISSUE #964: Legacy plugin application order breaks shadow detection logic updated for new id
### Expected and Results

I expect to be able to upgrade from 8.3.0 to 8.3.1 without errors.
Instead, I get [this](https://github.com/jpenilla/MiniMOTD/pull/226):
```
* What went wrong:
An exception occurred applying plugin request [id: 'com.gradleup.shadow']
> Failed to apply plugin class 'com.github.jengelman.gradle.plugins.shadow.legacy.LegacyShadowPlugin'.
   > Task with name 'shadowJar' not found in project ':minimotd-fabric'.
```
This is because my plugin `quiet-loom` is checking for all 3 shadow IDs: https://github.com/jpenilla/quiet-loom/blob/071311814f91e5b1409d5008976968da2dc60180/patches/architectury-loom/0006-Detect-shadowJar.patch

I think this could be fixed by applying the empty legacy plugin *after* plugins that register the shadowJar task.

### Related environent and versions

_No response_

### Reproduction steps

Have logic expecting `shadowJar` task to be registered inside of a `plugins.withId` action for the legacy plugin id.

### Anything else?

_No response_

jpenilla:
fixes #964 using the solution suggested by @ljacomet at https://github.com/GradleUp/shadow/issues/964#issuecomment-2345532565

Tested to fix the issue (bring behavior back to how it was before 8.3.1/with the old plugin), and to give an error when both the current and old plugin are on the classpath.

Something else that came to mind regarding the legacy plugin is that if someone puts the new plugin on the classpath, but then applies the legacy plugin id, nothing will happen, which could be confusing. But resolving that is likely better addressed in a separate PR (if at all).

```
> Could not resolve all files for configuration ':build-logic:compileClasspath'.
   > Could not resolve com.gradleup.shadow:shadow-gradle-plugin:8.3.3-SNAPSHOT.
     Required by:
         project :build-logic
      > Module 'com.gradleup.shadow:shadow-gradle-plugin' has been rejected:
           Cannot select module with conflict on capability 'com.github.johnrengelman:shadow:8.3.3-SNAPSHOT' also provided by [com.github.johnrengelman:shadow:8.1.1(apiElements)]
   > Could not resolve com.github.johnrengelman:shadow:8.1.1.
     Required by:
         project :build-logic
      > Module 'com.github.johnrengelman:shadow' has been rejected:
           Cannot select module with conflict on capability 'com.github.johnrengelman:shadow:8.1.1' also provided by [com.gradleup.shadow:shadow-gradle-plugin:8.3.3-SNAPSHOT(apiElements)]
```

---

### DIFF
--- a/build-logic/src/main/kotlin/shadow.convention.publish.gradle.kts
+++ b/build-logic/src/main/kotlin/shadow.convention.publish.gradle.kts
@@ -47,7 +47,7 @@ tasks.withType<Javadoc>().configureEach {
 }
 
 configurations {
-  sequenceOf(
+  listOf(
     apiElements,
     runtimeElements,
     named("javadocElements"),
--- a/build-logic/src/main/kotlin/shadow.convention.publish.gradle.kts
+++ b/build-logic/src/main/kotlin/shadow.convention.publish.gradle.kts
@@ -45,3 +45,39 @@ tasks.withType<Javadoc>().configureEach {
     it.addStringOption("Xdoclint:none", "-quiet")
   }
 }
+
+configurations {
+  sequenceOf(
+    apiElements,
+    runtimeElements,
+    named("javadocElements"),
+    named("sourcesElements"),
+  ).forEach {
+    it.configure {
+      attributes {
+        // We have dependencies that require Java 11 (jdependency)
+        attribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, 11)
+      }
+
+      outgoing {
+        // Main/current capability
+        capability("com.gradleup.shadow:shadow-gradle-plugin:$version")
+
+        // Historical capabilities
+        capability("io.github.goooler.shadow:shadow-gradle-plugin:$version")
+        capability("com.github.johnrengelman:shadow:$version")
+        capability("gradle.plugin.com.github.jengelman.gradle.plugins:shadow:$version")
+        capability("gradle.plugin.com.github.johnrengelman:shadow:$version")
+        capability("com.github.jengelman.gradle.plugins:shadow:$version")
+      }
+    }
+  }
+}
+
+publishing.publications.withType<MavenPublication>().configureEach {
+  // We don't care about capabilities being unmappable to Maven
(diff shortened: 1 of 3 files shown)


build-logic/src/main/kotlin/shadow.convention.publish.gradle.kts
```diff
@@ -45,3 +45,34 @@ tasks.withType<Javadoc>().configureEach {
     it.addStringOption("Xdoclint:none", "-quiet")
   }
 }
+
+configurations {
+  sequenceOf(
```
Goooler, line 50:
```suggestion
  listOf(
```

src/docs/changes/README.md
```diff
@@ -3,6 +3,7 @@
 
 ## [Unreleased]
 
+- Apply legacy plugin last, and declare capabilities for old plugins, fixes [#964](https://github.com/GradleUp/shadow/issues/964) ([#991](https://github.com/GradleUp/shadow/pull/991))
```
Goooler, line 6:
```suggestion
- Apply legacy plugin last, and declare capabilities for old plugins, fixes [#964](https://github.com/GradleUp/shadow/issues/964). ([#991](https://github.com/GradleUp/shadow/pull/991))
```

Goooler (approved):

Goooler:
CC @ljacomet

ljacomet (approved):
Sorry, I said I would attempt this and in the end did not find the time.

This LGTM!

The one thing I wondered was about the capability addition: should it be done in a patch release or in a minor?
My preference would be to do it in a minor as it might end up breaking builds (for good reasons) while the fix to the order of application fits perfectly a patch release.

ljacomet:
And thank you @jpenilla for taking care of fixing my mistake!

Goooler:
I'm planning to release this in 8.3.4, WDYT?

Issue #907, opened 2024-02-20
Plugin is incompatible with isolated projects feature


erdi:
Please check the [User Guide](http://imperceptiblethoughts.com/shadow) before submitting "how do I do 'x'?" questions!

### Shadow Version

8.1.1

### Gradle Version

8.5

### Expected Behavior

Plugin does not violate [project isolation](https://gradle.github.io/configuration-cache/) constraints, no errors are raised when a project to which the plugin is applied which also enables project isolation is being imported into IntelliJ.

### Actual Behavior

An error is thrown when a project to which the plugin is applied which also enables project isolation is being imported into IntelliJ. The following stacktrace is present in the configuration cache report generated upon the error:
```
org.gradle.api.InvalidUserCodeException: Cannot access project ':' from project ':services'
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGetProperty(AbstractCallSite.java:329)
	at com.github.jengelman.gradle.plugins.shadow.ShadowPlugin$_apply_closure1.doCall$original(ShadowPlugin.groovy:24)
	at com.github.jengelman.gradle.plugins.shadow.ShadowPlugin$_apply_closure1.doCall(ShadowPlugin.groovy)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:139)
	at com.github.jengelman.gradle.plugins.shadow.ShadowPlugin.apply(ShadowPlugin.groovy:15)
	at com.github.jengelman.gradle.plugins.shadow.ShadowPlugin.apply(ShadowPlugin.groovy)
	at org.gradle.api.internal.plugins.DefaultPluginContainer.apply(DefaultPluginContainer.java:70)
	....
```

That's because [the plugin calls `getPlugins()` on the `rootProject` instance from sub projects](https://github.com/johnrengelman/shadow/blob/2f94a2558023b208ffb840471281703cb2c9b222/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowPlugin.groovy#L24) which is not compatible with isolated projects.

[...]

PR #993, opened 2024-10-01
Disable Develocity integration by default

liutikas:
Configuring develocity build scan additional detail publishing breaks Gradle project isolation https://github.com/GradleUp/shadow/issues/907

This change adds a new property com.gradleup.shadow.enableDevelocityIntegration that defaults to true, however it allows folks to set it to false and make shadow plugin usage project isolation safe.

---

### DIFF
--- a/src/docs/changes/README.md
+++ b/src/docs/changes/README.md
@@ -3,6 +3,10 @@
 
 ## [Unreleased]
 
+**Changed**
+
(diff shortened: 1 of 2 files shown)


Goooler:
I'm considering removing develocity integration, WDYT?

3flex:
Or, merge this but default to disabling the integration, and add a deprecation warning if it's enabled.

If nobody complains then remove it in a couple of months.

Goooler (approved):

PR #1001, opened 2024-10-14
Fix documentation of custom tasks

quijote:
---

Adapts the docs at https://gradleup.com/shadow/custom-tasks/.

Using `testImplementation` does not work anymore with latest Gradle versions. It produces the following error:
```
Could not determine the dependencies of task ':testJar'.
> Resolving dependency configuration 'testImplementation' is not allowed as it is defined as 'canBeResolved=false'.
  Instead, a resolvable ('canBeResolved=true') dependency configuration that extends 'testImplementation' should be resolved.
```

Instead, [Gradle docs](https://docs.gradle.org/current/userguide/java_plugin.html) hints that `testRuntimeClasspath` is a resolvable configuration.

### DIFF
--- a/src/docs/custom-tasks/README.md
+++ b/src/docs/custom-tasks/README.md
@@ -12,11 +12,11 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 task testJar(type: ShadowJar) {
   archiveClassifier.set("tests")
   from sourceSets.test.output
-  configurations = [project.configurations.testImplementation]
+  configurations = [project.configurations.testRuntimeClasspath]
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/docs/.vuepress/config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   ga: "UA-321220-4",
   themeConfig: {
     repo: "GradleUp/shadow",
+    docsBranch: 'main',
```
Goooler, line 7: Can you send this change to a new PR?

quijote: https://github.com/GradleUp/shadow/pull/1002

Goooler (approved):

PR #1012, opened 2024-11-05
Finish Kotlin migration

Goooler:
Closes #980.

---

- [x] [CHANGELOG](src/docs/changes/README.md)'s "Unreleased" section has been updated, if applicable.

PR #1021, opened 2024-11-18
Kotlin migration part 4

Goooler:
Refs #1012.


api/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/MinimizeDependencyFilter.kt
```diff
@@ -0,0 +1,30 @@
...
+        includedDependencies: MutableSet<ResolvedDependency>,
+        excludedDependencies: MutableSet<ResolvedDependency>,
+    ) {
+        dependencies.forEach {
+            if (it.isIncluded() && !isParentExcluded(excludedDependencies, it)) {
+                includedDependencies.add(it)
+            } else {
+                excludedDependencies.add(it)
+            }
+            resolve(it.children, includedDependencies, excludedDependencies)
```
Goooler, line 20: This was wrongly converted!

Goooler: #1611

PR #1028, opened 2024-11-20
Kotlin migration: ShadowCopyAction

Goooler:
Refs #1012.


api/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.kt
```diff
@@ -0,0 +1,447 @@
...
+                        stats,
+                    ),
+                )
+                processTransformers(outputStream)
+            }
+        } catch (e: IOException) {
+            throw Zip64RequiredException(
+                "${e.cause?.message}\n\nTo build this archive, please enable the zip64 extension.\n" +
+                    "See: ${documentationRegistry.getDslRefForProperty(Zip::class.java, "zip64")}",
+            )
```
Goooler, line 123: We shouldn't rethrow all `IOException`s to `Zip64RequiredException`s, fixing in #1147.

PR #1036, opened 2024-11-21
Migrate transformers to using lazy properties

Goooler:
---


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.kt
```diff
@@ -156,22 +170,25 @@ public open class PropertiesFileTransformer : Transformer {
...
     val props = CleanProperties()
     // InputStream closed by caller, so we don't do it here.
-    props.load(inputStream.reader(_charset))
+    props.load(inputStream.reader(charset))
     return transformKeys(props)
   }
 
   private fun transformKeys(properties: Properties): CleanProperties {
-    if (keyTransformer === IDENTITY) {
+    if (keyTransformer == IDENTITY) {
```
Goooler, line 178: The new `keyTransformer` needs to be `get` before check equals.

Goooler: Fixing in #1204.

PR #1044, opened 2024-11-26
Migrate `ShadowJar` to using lazy properties

Goooler:
Refs #1036.

---

### DIFF
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
@@ -240,7 +240,7 @@ public abstract class ShadowJar :
       documentationRegistry,
       metadataCharset,
       transformers.get(),
-      allRelocators,
+      relocators.get() + packageRelocators,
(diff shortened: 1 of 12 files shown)


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -326,22 +268,21 @@ public abstract class ShadowJar :
...
+
+      val prefix = relocationPrefix.get()
+      // Must cast configurations to List<FileCollection> to fix type mismatch in runtime.
+      return (configurations.get() as List<FileCollection>).flatMap { configuration ->
+        configuration.files.flatMap { file ->
+          JarFile(file).use { jarFile ->
+            jarFile.entries().toList()
+              .filter { it.name.endsWith(".class") && it.name != "module-info.class" }
+              .map { it.name.substringBeforeLast('/').replace('/', '.') }
+              .map { SimpleRelocator(it, "$prefix.$it") }
```
Goooler, line 283: This is a regression, we have to reduce duplicates before creating relocators, fixed in #1271.

PR #1060, opened 2024-11-30
Migrate unit tests to Kotlin

Goooler:


Goooler:
The real change is https://github.com/GradleUp/shadow/compare/a25f017..89bd766

PR #1079, opened 2024-12-02
Only expose `includes` and `excludes` as inputs in `SimpleRelocator`

Goooler:
- Reworks #1047.
- Refs https://github.com/GradleUp/shadow/pull/524#discussion_r1865187155.

---

### DIFF
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.kt
@@ -2,8 +2,6 @@ package com.github.jengelman.gradle.plugins.shadow.relocation
 
 import java.util.regex.Pattern
 import org.codehaus.plexus.util.SelectorUtils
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/relocation/SimpleRelocator.kt
```diff
@@ -14,80 +15,62 @@ import org.gradle.api.tasks.Optional
...
+  private val pattern: String
+  private val pathPattern: String
+  private val shadedPattern: String
+  private val shadedPathPattern: String
+
+  @get:Input
+  public val includes: ListProperty<String> = objectFactory.listProperty(String::class.java)
+
+  @get:Input
+  public val excludes: ListProperty<String> = objectFactory.listProperty(String::class.java)
```
Goooler, line 34: They should be `SetProperty`s.

Goooler: Fixing in #1092.

Issue #852, opened 2023-04-07
Inconsistency between Jar Manifest class-path and dependencies


jbartok:
While investigating a [Gradle issue](https://github.com/gradle/gradle/issues/23928) when using the [Plugin Publish plugin](https://plugins.gradle.org/plugin/com.gradle.plugin-publish) together with the Shadow Plugin, we found an inconsistency in the Shadow plugin's code.

[Our solution](https://github.com/gradle/gradlecom/pull/1052) for it is just a workaround, we believe it would be good to fix the root cause, which is that while the `gradleApi()` is [excluded as a dependency](https://github.com/johnrengelman/shadow/blob/d530cff65c086e1cdd666bf3b12663949ae7ffcc/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy#L79) it is not excluded when the [Jar Manifest class-path is being built](https://github.com/johnrengelman/shadow/blob/d530cff65c086e1cdd666bf3b12663949ae7ffcc/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy#L64). It should instead work [as the `java-gradle-plugin` works](https://github.com/gradle/gradle/blob/50e3e8b110da3c5d846abd4d66776ef277d2a940/subprojects/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/JavaGradlePluginPlugin.java#L207).

### Shadow Version

Latest.

### Gradle Version

Not relevant.

### Content of Shadow JAR (`jar tf <jar file>` - post link to GIST if too long)

See linked Gradle issue above. The problem is observed in the Manifest file of the Jar.

johnrengelman:
Thanks @jbartok. I'm looking at the `java-gradle-plugin` and I'm wondering what the best way for me to implement the functionality [here](https://github.com/gradle/gradle/blob/50e3e8b110da3c5d846abd4d66776ef277d2a940/subprojects/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/JavaGradlePluginPlugin.java#L216-L222) without using internal APIs.
Also, it's appears your first solution is a link into the close-source plugin-publish code, so I can't see that...Not sure it matters to see the workaround though.

johnrengelman:
I really would love if these Gradle specific dependencies weren't special cases from normal dependencies (similar to https://github.com/gradle/gradle/issues/24018).

johnrengelman:
[...]

PR #1149, opened 2025-01-12
Check `Class-Path` in shadowed Gradle plugin jar

Goooler:
Refs #852.

### DIFF
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPluginTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPluginTest.kt
@@ -3,10 +3,13 @@ package com.github.jengelman.gradle.plugins.shadow
 import assertk.all
 import assertk.assertThat
 import assertk.assertions.contains
+import assertk.assertions.doesNotContain
 import assertk.assertions.isEqualTo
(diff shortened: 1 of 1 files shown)


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPluginTest.kt
```diff
@@ -765,8 +769,15 @@ class ShadowPluginTest : BasePluginTest() {
...
 
-    val entries = outputShadowJar.use { it.entries().toList() }
-    assertThat(entries.count { it.name.endsWith(".class") }).isEqualTo(1)
+    assertThat(outputShadowJar).useAll {
+      transform { actual -> actual.entries().toList().map { it.name }.filter { it.endsWith(".class") } }
+        .single().isEqualTo("my/plugin/MyPlugin.class")
+      transform { it.manifest.mainAttributes.keys }.all {
+        isNotEmpty()
+        // Doesn't contain Gradle classes.
+        doesNotContain("Class-Path")
```
Goooler, line 778: This may not work properly. The key set is `Set<Object>` instead of `Set<String>`.

Goooler: Fixing in #1161.

PR #1150, opened 2025-01-12
Show how to publish the shadowed jar only

### LINKED ISSUE #986: Publishing a shaded JAR
### Describe your use-case which is not covered by existing documentation.

I need to shade a specific library that does not maintain binary compatibility, I thus want to publish a JAR that has shadowed and relocated a specific library, while all other libraries are exposes in the POM like normal.

This worked fine until 8.3.1 when #904 was merged. 

The issue is that the shadowJar task uses the `all` classifier by default which isn't supported by non Gradle clients. You can manually set the classifier to `null` or `''` but then it conflicts with the `jar` task.

It would be great to call this specific use case out in the docs as it's quite common.

### Reference any relevant documentation, other materials or issues/pull requests that can be used for inspiration.

My current solution can be found in my [auto-dagger](https://github.com/ansman/auto-dagger/blob/main/build-logic/src/main/kotlin/library.publishing.gradle.kts) project. 

Here is the relevant snippet:
```kotlin
pluginManager.withPlugin("com.gradleup.shadow") {
    val compileShaded: Configuration by configurations.creating
    configurations.named("compileOnly") { extendsFrom(compileShaded) }
    configurations.named("testRuntimeOnly") { extendsFrom(compileShaded) }
    // Since we use a custom configuration we make the shadow configuration extend the implementation
    // configuration to correctly include dependencies in the POM
    configurations.named("shadow") { extendsFrom(configurations["implementation"]) }

    // Since we change the classifier of the shadowJar we need to disable the default jar task or we'll get two
    // artifacts that have the same classifier
    tasks.named<Jar>("jar") {
        archiveClassifier.set("ignored")
    }

    tasks.named<ShadowJar>("shadowJar") {
        archiveClassifier.set("")
        configurations = listOf(compileShaded)
        isEnableRelocation = true
        relocationPrefix = "se.ansman.dagger.auto${project.path.replace(':', '.').replace('-', '_')}"
        mergeServiceFiles()
    }

    pluginManager.withPlugin("org.gradle.java-test-fixtures") {
        configurations.named("testFixturesImplementation") { extendsFrom(compileShaded) }
        configurations.named("testFixturesImplementation") { extendsFrom(configurations["implementation"]) }
        configurations.named("testFixturesApi") { extendsFrom(configurations["api"]) }
    }
}
```

The main downside of this is that other projects that depend on this would have to depend on the `shadow` configuration which in my case isn't an issue since none of these dependant modules are published but I can imagine it would be an issue if they were.

### Are you interested in contributing to the documentation?

_No response_

Goooler:
Closes #986.

### DIFF
--- a/src/docs/publishing/README.md
+++ b/src/docs/publishing/README.md
@@ -46,7 +46,7 @@ configuring publishing. If this behavior is not desirable, then publishing **mus
 be manually configured.
 
 
-# Publish the Shadowed JAR instead of the Original JAR
+## Publish the Shadowed JAR instead of the Original JAR
 
 You may want to publish the shadowed JAR instead of the original JAR. This can be done by trimming 
 the `archiveClassifier` of the shadowed JAR like the following:
--- a/src/docs/publishing/README.md
+++ b/src/docs/publishing/README.md
@@ -47,7 +47,7 @@ be manually configured.
 
 # Publish the Shadowed JAR instead of the Original JAR
 
-You may want to publish the shadowed JAR instead of the original JAR. This can be done by trim 
+You may want to publish the shadowed JAR instead of the original JAR. This can be done by trimming 
 the `archiveClassifier` of the shadowed JAR like the following:
 
 ```groovy
--- a/src/docs/publishing/README.md
+++ b/src/docs/publishing/README.md
@@ -59,7 +59,7 @@ group = 'shadow'
 version = '1.0'
 
 dependencies {
-  // This will be bundled in the shadowed JAR.
+  // This will be bundled in the shadowed JAR and not declared in the POM.
   implementation 'some:a:1.0'
   // This will be excluded from the shadowed JAR but declared as a runtime dependency in `META-INF/MANIFEST.MF`
   // file's `Class-Path` entry, and also in the POM file.
--- a/src/docs/publishing/README.md
+++ b/src/docs/publishing/README.md
@@ -44,3 +44,44 @@ configuration of the POM file.
 This automatic configuration occurs _only_ when using the above methods for
 configuring publishing. If this behavior is not desirable, then publishing **must**
 be manually configured.
+
+# Publish the Shadowed JAR instead of the Original JAR
+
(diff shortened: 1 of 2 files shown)


src/docs/publishing/README.md
```diff
@@ -44,3 +44,44 @@ configuration of the POM file.
 This automatic configuration occurs _only_ when using the above methods for
 configuring publishing. If this behavior is not desirable, then publishing **must**
 be manually configured.
+
+# Publish the Shadowed JAR instead of the Original JAR
+
+You may want to publish the shadowed JAR instead of the original JAR. This can be done by trim
```
Goooler, line 50:
```suggestion
You may want to publish the shadowed JAR instead of the original JAR. This can be done by trimming 
```

src/docs/publishing/README.md
```diff
@@ -44,3 +44,44 @@ configuration of the POM file.
...
+```groovy
+apply plugin: 'java'
+apply plugin: 'maven-publish'
+apply plugin: 'com.gradleup.shadow'
+
+group = 'shadow'
+version = '1.0'
+
+dependencies {
+  // This will be bundled in the shadowed JAR.
```
Goooler, line 62:
```suggestion
  // This will be bundled in the shadowed JAR and not declared in the POM.
```

src/docs/publishing/README.md
```diff
@@ -44,3 +44,44 @@ configuration of the POM file.
 This automatic configuration occurs _only_ when using the above methods for
 configuring publishing. If this behavior is not desirable, then publishing **must**
 be manually configured.
+
```
Goooler, line 47:
```suggestion


```

src/docs/publishing/README.md
```diff
@@ -44,3 +44,45 @@ configuration of the POM file.
...
+
+# Publish the Shadowed JAR instead of the Original JAR
```
Goooler, line 49:
```suggestion
## Publish the Shadowed JAR instead of the Original JAR
```

PR #1156, opened 2025-01-14
Test publishing shadowed Gradle plugin

Goooler:
Follow up 

- #1143.
- #1150.


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/PublishingTest.kt
```diff
@@ -84,20 +87,51 @@ class PublishingTest : BasePluginTest() {
...
+            from components.shadow
+          }
+        """.trimIndent(),
+      ),
+    )
+
+    publish()
+
+    assertShadowJarCommon(repoJarPath("my/plugin/maven/1.0/maven-1.0.jar"))
+    assertPomCommon(repoPath("my/plugin/maven/1.0/maven-1.0.pom"))
```
Goooler, line 123: `val dependency = pom.dependencies.single()` is flaky!

Goooler: 8e1d10b2dc0b0bb43cb9c394d1a5eb828366ae40

PR #1157, opened 2025-01-14
Enable Isolated Projects for project self

Goooler:


Goooler:
Seems nothing worthy of the single project.

Issue #891, opened 2023-10-12
`args` in `runShadow` are not added into start scripts


injecteer:
Please check the [User Guide](http://imperceptiblethoughts.com/shadow) before submitting "how do I do 'x'?" questions!

### Expected Behavior
The `args` config parameter should be used along with `applicationDefaultJvmArgs` and be written in  `scriptsShadow/*` scripts

### Actual Behavior
`args` are not to be found in run-scripts

### Gradle Build Script(s)

```
plugins {
  id 'application'
  id 'com.github.johnrengelman.shadow' version '8.1.1'
}

applicationDefaultJvmArgs = [
  '-Djava.net.preferIPv4Stack=true', 
  '-Dfile.encoding=UTF-8',
  // hazelcast args for Java 9+
  '--add-modules', 'java.se', 
  '--add-exports', 'java.base/jdk.internal.ref=ALL-UNNAMED', 
  '--add-opens', 'java.base/java.lang=ALL-UNNAMED', 
  '--add-opens', 'java.base/java.nio=ALL-UNNAMED', 
  '--add-opens', 'java.base/sun.nio.ch=ALL-UNNAMED', 
  '--add-opens', 'java.management/sun.management=ALL-UNNAMED', 
  '--add-opens', 'jdk.management/com.sun.management.internal=ALL-UNNAMED'
]

runShadow {
  args '-Dvertxweb.environment=prod', '-cluster'
}
```

### Content of Script files
[scriptsShadow.zip](https://github.com/johnrengelman/shadow/files/12879635/scriptsShadow.zip)

Goooler:
`args` in `runShadow` can't be added into start scripts, you should call them like:

```sh
./unixStartScript.sh '-Dvertxweb.environment=prod', '-cluster'
```

PR #1195, opened 2025-01-26
Verify `DEFAULT_JVM_OPTS` in start scripts

Goooler:
Refs #891.

### DIFF
--- a/src/docs/application-plugin/README.md
+++ b/src/docs/application-plugin/README.md
@@ -68,7 +68,7 @@ plugins {
 
 application {
   mainClass = 'myapp.Main'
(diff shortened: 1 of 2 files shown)


src/docs/application-plugin/README.md
```diff
@@ -68,6 +68,8 @@ plugins {
 
 application {
   mainClass = 'myapp.Main'
+  // Optionally, you can add extra JVM arguments to the start scripts.
```
Goooler, line 71:
```suggestion
  // Optionally, you can add default JVM arguments to the start scripts like this:
```

PR #1197, opened 2025-01-26
Replace `JavaJarExec` with `JavaExec`

Goooler:
---


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.kt
```diff
@@ -45,18 +45,18 @@ public abstract class ShadowApplicationPlugin : Plugin<Project> {
...
+    project.tasks.register(SHADOW_RUN_TASK_NAME, JavaExec::class.java) {
       val install = project.tasks.named(SHADOW_INSTALL_TASK_NAME, Sync::class.java)
+      val jarFile = install.zip(shadowJar) { i, s ->
+        i.destinationDir.resolve("lib/${s.archiveFile.get().asFile.name}")
+      }
+
       it.dependsOn(install)
-      it.mainClass.set("-jar")
+      it.classpath(jarFile)
+      it.mainClass.set(javaApplication.mainClass)
```
Goooler, line 56: Set `mainClass` from `javaApplication.mainClass` instead of `-jar` and make sure the `jarFile` is in `classpath`, so that we can get rid of the args appending logic in 

https://github.com/GradleUp/shadow/blob/4850db3f3a9c5755d9d38a43b9337cea0838c4ba/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/JavaJarExec.kt#L14-L19

Goooler:
@ljacomet Can you take a review for this? I noticed you did some adoption in #576.

Goooler:
The newly added test in #1214 passed both before and after this change, so everything looks good.

Feel free to have a post review.

PR #1203, opened 2025-01-27
Get rid of `Closure` in public APIs

Goooler:
Address #1088.

---

- [ ] [CHANGELOG](https://github.com/GradleUp/shadow/blob/main/src/docs/changes/README.md)'s "Unreleased" section has been updated, if applicable.


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.kt
```diff
@@ -124,7 +123,7 @@ public open class PropertiesFileTransformer @Inject constructor(
   public open val charsetName: Property<String> = objectFactory.property(Charsets.ISO_8859_1.name())
 
   @get:Internal
-  public open val keyTransformer: Property<Closure<String>> = objectFactory.property(IDENTITY)
+  public open val keyTransformer: Property<JavaFunction<String, String>> = objectFactory.property(IDENTITY)
```
Goooler, line 126:
```
* What went wrong:	
A problem occurred configuring root project 'shadow'.	
> Could not create task ':shadowJar'.	
   > Cannot set the value of property 'keyTransformer' of type java.util.function.Function using an instance of type build_6653fgw173xqzj0qt02ylgus0$_run_closure1$_closure2$_closure3.	
```

`Closure` type can't be accepted by `set` call.

https://scans.gradle.com/s/jobkdojvttqxw/tests/overview?outcome=FAILED

PR #1210, opened 2025-01-28
More cacheable `Transformer`s

Goooler:
Follow up 

- #524.
- #1036.

---


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheLicenseResourceTransformer.kt
Goooler, line 1: No `@Input` in this transformer.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/ComponentsXmlResourceTransformer.kt
Goooler, line 1: No `@Input` in this transformer too.

PR #1233, opened 2025-02-06
Refactor file visiting logic in StreamAction

### LINKED ISSUE #53: Relocation moves classes to new package, but keeps old package in the JAR.
When I relocate the guava package in [my test project](https://github.com/Minecrell/GradleTest) to another package, it moves the classes correctly but if I look into the JAR the folder of the original package still exists.

``` groovy
relocate 'com.google.common', 'net.minecrell.gradletest.com.google.common'
```

Guava has been moved to the correct package but there is still the `com` folder with all of the sub directories as well.

### LINKED ISSUE #57: Lingering package name & folders for relocations in source project.
From @minecrell - https://github.com/johnrengelman/shadow/issues/53

Just tested the new version 1.0.0 and it doesn't generate the old folders of the dependencies anymore. However, for some reason in the JAR file is an empty folder with the name of the file. (In my case GradleTest-1.0-all). This is kinda strange but I don't really mind, but maybe it is just a small mistake that is easily fixable?
Relocations of packages within the source files still generate their old folders in the JAR but as of I don't use that, it's not important to fix (for me).

### LINKED ISSUE #111: Excluding a jar from being unzipped
I've got a `.jar` file in my tree that I want to get included into the shadowjar, but not unzipped into the shadowjar. Right now it's unclear to me how I might have the file _included_ but not unpacked.

Guidance is much appreciated :D

### LINKED ISSUE #475: Duplicates Strategy of Jar task is ignored for files of dependencies
Given `ProjectA` and `ProjectB`
`ProjectB` depends on `ProjectA`
`ProjectB` has the shadow jar plugin applied
Both projects have `com.example.ClassX`
duplicatesStrategy is configured for the `shadowJar` task as shown below.

### Shadow Version
5.0.0

### Gradle Version
5.3.1

### Actual Behavior
running `shadowJar` task is successful and only the `ProjectB`'s `ClassX` is in the jar (I think, haven't checked if the zip has multiple entries with same path)

### Expected Behavior
Either documentation for duplicatesStrategy behavior, or have same behavior as the specified duplicatesStrategy (Fail in this case)

### Gradle Build Script(s)
``` gradle
shadowJar {
    duplicatesStrategy = DuplicatesStrategy.FAIL
}
```

### LINKED ISSUE #488: DuplicatesStrategy.FAIL doesn't work with transform
I have something along the lines of:

```gradle
task foo(type: ShadowJar, dependsOn: fooClasses) {
	from sourceSets.foo.output

	configurations = [project.configurations.fooRuntimeClasspath]

	archiveBaseName = 'foo'
	zip64 = true

	mergeServiceFiles()
	mergeGroovyExtensionModules()
	transform(AppendingTransformer) {
		resource = 'META-INF/extensions.idx'
	}

	duplicatesStrategy = DuplicatesStrategy.FAIL
}
```

There are multiple `META-INF/extensions.idx` files which, when `duplicatesStrategy` is NOT `DuplicatesStrategy.FAIL`, get merged correctly but when it IS `DuplicatesStrategy.FAIL` - it blows up with `org.gradle.api.file.DuplicateFileCopyingException: Encountered duplicate path "META-INF/extensions.idx" during copy operation configured with DuplicatesStrategy.FAIL`. Am I doing something wrong or is this a bug? I think that having specified that I want those files appended means that it should not blow up with that error, right?

### LINKED ISSUE #766: Add unzipped dependency jars to fat jar

### Shadow Version
7.1.2
### Gradle Version
7.4.2
### Expected Behavior
Need adding jar files of dependencies, not unzipped classes
### Actual Behavior
Added unzipped class files from dependencies to shadowJar
### Gradle Build Script(s)
default shadowJar in Idea

### LINKED ISSUE #1223: Duplicates from jars are not handled per duplicatesStrategy
### Expected and Results

`ShadowJar` task supports re-embedding of contents of other jars using the `shadowTask.from(jarToEmbed)` API. However, duplicates encountered across the jars always get excluded, regardless of the `duplicatesStrategy` configuration. The only duplicates the `ShadowJar` recognizes are the duplicates among the copied files.

The expected behavior is that `ShadowJar` treats duplicates per `duplicationStrategy` regardless of whether they come from.

### Related environent and versions

_No response_

### Reproduction steps

Consider the following `ShadowJar` task configuration:

```kotlin
import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
import java.util.zip.ZipInputStream

plugins {
    id("com.github.johnrengelman.shadow") version "8.1.1"
}

fun makeFoo(value: String): TaskProvider<*> = tasks.register(value) {
    val out = layout.buildDirectory.file("${value}/foo")
    outputs.file(out)
    doLast {
        out.get().asFile.writeText(value)
    }
}

val foo1 = makeFoo("1")
val jar1 = tasks.register<Jar>("jar1") {
    from(foo1)
    archiveBaseName.set("jar1")
    destinationDirectory.set(layout.buildDirectory.dir("jar1"))
}
val foo2 = makeFoo("2")
val jar2 = tasks.register<Jar>("jar2") {
    from(foo2)
    archiveBaseName.set("jar2")
    destinationDirectory.set(layout.buildDirectory.dir("jar2"))
}

val foo3 = makeFoo("3")
val foo4 = makeFoo("4")

val shadow = tasks.register<ShadowJar>("shadow") {
    // this does nothing in terms of duplicates across embedded jars
    duplicatesStrategy = DuplicatesStrategy.FAIL
    archiveBaseName.set("shadow")
    from(
        jar1,
        jar2, // duplicate from this jar gets silently excluded
        foo3, // this is also silently ignored
        foo4, // but this duplicate is suddenly recognized as a duplicate
    )
    val output = layout.buildDirectory.dir("shadow")
    destinationDirectory.set(output)

    doLast {
        println(
            ZipInputStream(
                output.get().asFile.resolve("shadow.jar").inputStream()
            ).use {
                generateSequence { it.nextEntry }.first { it.name == "foo" }
                it.readBytes().decodeToString()
            }
        )
    }
}
```

Duplication among `foo3` and `foo4` is recognized, but the duplicate file `foo` between `jar1`, `jar2` and `foo3` for example is silently excluded.

### Anything else?

_No response_

Goooler:
- Closes #53.
- Closes #57.
- Closes #111.
- Closes #475.
- Closes #488.
- Closes #766.
- Closes #1223.
- Closes https://github.com/Goooler/shadow/issues/92.

---


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/AppendingTransformerTest.kt
```diff
@@ -74,7 +74,15 @@ class AppendingTransformerTest : BaseTransformerTest() {
       block1 + System.lineSeparator() + block2
     }
 
-    projectScriptPath.appendText(config)
+    projectScriptPath.appendText(config + System.lineSeparator())
+    projectScriptPath.appendText(
+      """
+      $shadowJar {
+        // This should fail the build because of the duplicates.
+        duplicatesStrategy = DuplicatesStrategy.FAIL
```
Goooler, line 82: This works for

```kt
val fatJar by tasks.registering(Jar::class) {
  archiveClassifier = "all"
  duplicatesStrategy = DuplicatesStrategy.FAIL

  from(sourceSets.main.map { it.output })
  from(configurations.runtimeClasspath.map { it.asFileTree.files.map(::zipTree) })
}
```

src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/AppendingTransformerTest.kt
```diff
@@ -100,6 +104,31 @@ class AppendingTransformerTest : BaseTransformerTest() {
...
+          // This should fail the build because of the duplicates.
+          duplicatesStrategy = DuplicatesStrategy.FAIL
+        }
+      """.trimIndent(),
+    )
+
+    val result = runWithFailure(shadowJarTask)
+
+    assertThat(result).all {
+      taskOutcomeEquals(shadowJarTask, FAILED)
```
Goooler, line 125: This works for a custom fat jar, see 388909da07b900df1f6c9b2e3889b800611950f9.

Goooler: As `ZipTree` inputs are handled by https://github.com/gradle/gradle/blob/df5bc230c57db70aa3f6909403e5f89d7efde531/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/copy/DuplicateHandlingCopyActionDecorator.java#L54-L65

PR #1234, opened 2025-02-06
Remove unnecessary `@Internal` in `ComponentsXmlResourceTransformer`

Goooler:
Refs https://github.com/apache/maven-shade-plugin/blob/3a844d0fa5ad7c283507a274ba1455c1207b9bd6/src/main/java/org/apache/maven/plugins/shade/resource/ComponentsXmlResourceTransformer.java#L142-L160.


Goooler:
Oooops, we still need it.

PR #1276, opened 2025-02-24
Shorten obtaining relative paths

Goooler:
> Returns the path of this file, relative to the root of the containing file tree. Always uses '/' as the hierarchy
  separator, regardless of platform file separator. Same as calling getRelativePath().getPathString().

Closes #1273.


Goooler:
Need things that depend on #1233.

https://scans.gradle.com/s/aekicvxwbblgg/tests/overview?outcome=FAILED

PR #1289, opened 2025-02-27
Remove `Named` from the parents of `Transformer`

Goooler:
It's unused.

---


wolfs:
This was used by Gradle when deciding on names for things in a list as nested input properties. So this should have been helpful when trying to do Build Scan comparisons between two builds to see what changed. You can see the code on the Gradle side here: https://github.com/gradle/gradle/blob/cb0c615fb8e3690971bb7f89ad80f58943360624/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/properties/annotations/AbstractTypeMetadataWalker.java#L158-L168

Goooler:
Do you mean we should revert this change?

wolfs:
Yes, that is my suggestion.

PR #1293, opened 2025-02-28
Try to use .github/workflows/.java-version

Goooler:


.github/workflows/ci.yml
```diff
@@ -13,14 +13,15 @@ jobs:
...
-        java: [ 17, 21, 23 ]
+        java: [ 17, 21 ]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
           java-version: ${{ matrix.java }}
+          java-version-file: .github/workflows/.java-version
```
Goooler, line 24: Seems we can't mix them.

> java-version: The Java version that is going to be set up. Takes a whole or [semver](https://github.com/actions/setup-java?tab=readme-ov-file#supported-version-syntax) Java version. If not specified, the action will expect java-version-file input to be specified.

https://github.com/actions/setup-java?tab=readme-ov-file#usage

PR #1343, opened 2025-03-13
Document handing `DuplicatesStrategy`

### LINKED ISSUE #520: Add docs for the usages of `DuplicatesStrategy`
Hi!

It would be very helpful to be able to exclude files but only if they come from dependencies, not project's. Some examples:
- **META-INF/LICENSE**
- **META-INF/NOTICE**
- **META-INF/versions/9/module-info.class**

If we exclude them with the current API, project's files will also get excluded from the final JAR, but that's very undesired.

Thank you!

Goooler:
Closes #520.

### DIFF
--- a/docs/configuration/merging/README.md
+++ b/docs/configuration/merging/README.md
@@ -358,7 +358,7 @@ It must be added using the [`transform`](https://gradleup.com/shadow/api/shadow/
 
 ## Handing Duplicates Strategy
 
-`ShaoowJar` is a subclass of 
+ShadowJar` is a subclass of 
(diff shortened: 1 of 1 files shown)


docs/configuration/merging/README.md
```diff
@@ -355,3 +355,44 @@ It must be added using the [`transform`](https://gradleup.com/shadow/api/shadow/
       }
     }
     ```
+
+## Handling Duplicates Strategy
+ShadowJar` is a subclass of
```
Goooler, line 360:
```suggestion

`ShadowJar` is a subclass of 
```

Issue #1377, opened 2025-03-28
Eagerly create jvm target for KMP


Goooler:
> Please don't call `jvm()`. Because this method declares a jvm target in KMP project. And users should do that explicitly. 
> More over, they can specify a custom name for their jvm target. 
> The correct API is to lazily iterate over targets collection. And check for KotlinJvmTarget instance.

```kt
kmpExtension.targets.configureEach { target ->
   if (target is KotlinJvmTarget) // do configuration here
}
```

_Originally posted by @antohaby in https://github.com/GradleUp/shadow/pull/1280#discussion_r2018890845_

PR #1379, opened 2025-03-28
Test `doNotCreateJvmTargetEagerly`

Goooler:
Refs #1377.


Goooler:
This doesn't work for main as expected, as the `jvm` target is created eagerly by 

https://github.com/GradleUp/shadow/blob/7d204ca3f32ebc9696fb5abf56b1e7f634774af3/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowKmpPlugin.kt#L19-L21

PR #1382, opened 2025-04-01
Replace composite deployment with the general workflow

Goooler:


Goooler:
Workflow dispatch works well in https://github.com/GradleUp/shadow/actions/runs/14185980866.

PR #1390, opened 2025-04-10
Update Jar task references

Goooler:


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -165,10 +164,6 @@ public abstract class ShadowJar :
   @Input // Trigger task executions after excludes changed.
   override fun getExcludes(): MutableSet<String> = super.getExcludes()
 
-  override fun manifest(configureClosure: Closure<*>): ShadowJar = apply {
-    super.manifest(configureClosure)
-  }
```
Goooler: This seems useless, and I don't want to expose `Closure` as the public API again.

PR #1397, opened 2025-04-13
Use dokka 2

Goooler:
https://kotlinlang.org/docs/dokka-migration.html


Goooler:
https://github.com/GradleUp/shadow/actions/runs/14426742489/job/40456659624

PR #1398, opened 2025-04-13
Eliminate convention plugin

Goooler:


build-logic/src/main/kotlin/shadow.convention.publish.gradle.kts
```diff
@@ -1,78 +0,0 @@
...
-dokka {
-  dokkaPublications.html {
-    outputDirectory = rootDir.resolve("docs/api")
-  }
-}
-
-java {
-  withSourcesJar()
-  withJavadocJar()
-}
```
Goooler, line 25: This is unnecessary after migrating the codebase to Kotlin.

build.gradle.kts
```diff
@@ -61,7 +105,6 @@ dependencies {
   implementation(libs.plexus.xml)
 
   testPluginClasspath(libs.foojayResolver)
-  testPluginClasspath(libs.pluginPublish)
```
Goooler, line 64: This should be unnecessary after updating to newer Gradle versions.

Goooler: Still required now, ac9f918c90b6a5e8f1379683c7253a3c95fb4a80

PR #1405, opened 2025-04-17
Test that string consts are relocated by default

Goooler:
Refs #1401.


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/RelocationTest.kt
```diff
@@ -589,12 +589,9 @@ class RelocationTest : BasePluginTest() {
       """
         package my;
         public class Main {
-          public static final String junit = "junit.framework.Test";
           public static void main(String[] args) {
-            System.out.println(getValue() + junit);
+            System.out.println("junit.framework.Test");
           }
-          // Use this method to force the compiler to not inline the string literal.
-          private static String getValue() { return "the value is "; }
```
Goooler: `getValue() + junit` has still been inlined, we should use a const directly.

PR #1422, opened 2025-05-03
Don't re-add suppressed Gradle API to `compileOnly` configuration

Jonathing:
As part of Gradle's experiments with publishing an external Gradle API (gradle/gradle#29483), they are beginning to look at suppressing the local Gradle API dependency that is usually used when the `java-gradle-plugin` plugin is applied. Because of this, though, the current strategy that the Shadow plugin uses to move it from `api` to `compileOnly` doesn't check if it exists to begin with. This PR aims to solve that problem. I've added tests as needed.

Gradle's Plugin Publish plugin has a similar issue which I've gone into detail in gradle/plugin-portal-requests#260.

---

### DIFF
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
@@ -5,6 +5,7 @@ import assertk.assertThat
 import assertk.assertions.contains
 import assertk.assertions.containsMatch
 import assertk.assertions.containsOnly
+import assertk.assertions.doesNotContain
 import assertk.assertions.isEqualTo
 import assertk.assertions.isFalse
 import assertk.assertions.isGreaterThan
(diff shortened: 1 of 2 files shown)


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
```diff
@@ -519,6 +522,44 @@ class JavaPluginTest : BasePluginTest() {
...
+
+    val outputCompileOnly = dependencies(COMPILE_ONLY_CONFIGURATION_NAME)
+    val outputApi = dependencies(API_CONFIGURATION_NAME)
+
+    // "unspecified" is the local Gradle API.
+    assertThat(outputCompileOnly).contains("unspecified")
+    assertThat(outputApi).doesNotContain("unspecified")
+  }
+
+  @ParameterizedTest
```
Goooler, line 544:
```suggestion
  @Issue(
    "https://github.com/GradleUp/shadow/issues/1422",
  )
  @ParameterizedTest
```

Goooler (approved):
LGTM!

I'll merge this after feedback from https://github.com/gradle/plugin-portal-requests/issues/260.

Jonathing:
Hell yeah

PR #1429, opened 2025-05-16
There's no need to run update-start-scripts job frequently

Goooler:
The start scripts are typically updated with Gradle releases.

Refs https://github.com/Goooler/agp-sources/blob/ac23064e6529e504d1dbabbd2cccec3c60bc4c2c/.github/workflows/ci.yml#L22-L45.


Goooler:
Kill https://github.com/GradleUp/shadow/actions/workflows/start-scripts.yml.

PR #1441, opened 2025-05-27
Allow using file trees of JARs together with the configuration cache

wolfs:
We are adding JARs from a directory that is built by a task to the `ShadowJar` by using `includedDependencies`. That doesn't seem to work when using the configuration cache, since the contents of the directory are evaluated at configuration time, causing them always to be empty.

This PR fixes that by doing the actual resolution as late as in the task action, not when the `includedZipTrees` provider is being stored in the configuration cache.

---


wolfs:
@Goooler Thank you for merging this so quickly! Any idea when you are going to do another release that includes the changes?

Goooler:
It will be out in a few hours.

PR #1491, opened 2025-06-27
Use Java 21 for the build (the target is still Java 11)

vlsi:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,9 +23,19 @@ version = providers.gradleProperty("VERSION_NAME").get()
 group = providers.gradleProperty("GROUP").get()
 description = providers.gradleProperty("POM_DESCRIPTION").get()
 
+// Release artifacts target Java 11; however, some tests require Java 17+
+val testJdk = providers.gradleProperty("testJdk").getOrElse("17")
+
 dokka {
   dokkaPublications.html {
     outputDirectory = rootDir.resolve("docs/api")
+  }
+}
+
+java {
+  toolchain {
+    // Use Java 21 for building the project (a recent LTS) to minimize the chances of bugs in the compiler
+    languageVersion = JavaLanguageVersion.of(21)
   }
 }
 
@@ -168,6 +178,11 @@ testing.suites {
     targets.configureEach {
       testTask {
         maxParallelForks = Runtime.getRuntime().availableProcessors()
+        javaLauncher.set(
+          javaToolchains.launcherFor {
+            languageVersion.set(JavaLanguageVersion.of(testJdk))
+          }
+        )
       }
     }
   }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,9 +20,13 @@ jobs:
       - uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
-          java-version: ${{ matrix.java }}
+          # Setup the selected Java, and configure Java 21 by default
+          java-version: |
+            ${{ matrix.java }}
+            21
       - uses: gradle/actions/setup-gradle@v4
-      - run: ./gradlew build
+      # We provision JDKs with GitHub Actions for caching purposes, so Gradle should rather fail in case JDK is not found
+      - run: ./gradlew -Porg.gradle.java.installations.auto-download=false -PtestJdk=${{ matrix.java }} build
 
   # There's no need to run the update task frequently; the start scripts are typically updated with Gradle releases.
   update-start-scripts:
@@ -61,7 +65,8 @@ jobs:
         with:
           cache-read-only: true
       # TODO: https://github.com/gradle/gradle/issues/22779
-      - run: ./gradlew publishToMavenCentral --no-configuration-cache
+      # We provision JDKs with GitHub Actions for caching purposes, so Gradle should rather fail in case JDK is not found
+      - run: ./gradlew -Porg.gradle.java.installations.auto-download=false publishToMavenCentral --no-configuration-cache
         env:
           ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.CENTRAL_PORTAL_USERNAME }}
           ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.CENTRAL_PORTAL_PASSWORD }}
--- a/docs/changes/README.md
+++ b/docs/changes/README.md
@@ -5,6 +5,7 @@
 **Changed**
 
 - Expose Ant as `compile` scope. ([#1488](https://github.com/GradleUp/shadow/pull/1488))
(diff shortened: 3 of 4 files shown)


build.gradle.kts
```diff
@@ -23,12 +23,22 @@ version = providers.gradleProperty("VERSION_NAME").get()
...
 dokka {
   dokkaPublications.html {
     outputDirectory = rootDir.resolve("docs/api")
   }
 }
 
+java {
+  toolchain {
+    // Use Java 21 for building the project (a recent LTS) to minimize the chances of bugs in the compiler
+    languageVersion = JavaLanguageVersion.of(21)
```
Goooler, line 38: Intended for not using toolchain, due to some reasons like https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/.

build.gradle.kts
```diff
@@ -168,6 +178,11 @@ testing.suites {
     targets.configureEach {
       testTask {
         maxParallelForks = Runtime.getRuntime().availableProcessors()
+        javaLauncher.set(
+          javaToolchains.launcherFor {
+            languageVersion.set(JavaLanguageVersion.of(testJdk))
+          }
```
Goooler, line 184: This defaults to `java.toolchain`, I guess you wanted to use `testJdk` in `testing` block.

build.gradle.kts
vlsi: Frankly, I find that blogpost seriously misleading. It has a clickbait title, however, the justification is missing (detailed analysis follows).

Note:
* I still want convincing you that this PR (use the latest LTS for the build) was a good idea, and my PR **agrees** with the blogpost you mentioned. See "[compile with the latest Java, but test through the lowest-supported Java](https://jakewharton.com/build-on-latest-java-test-through-lowest-java/)". Well, I do not fully understand what they say by "latest Java", and I would refrain from compiling with non-LTS versions as they are much less tested than LTS ones
* You are free to agree or disagree (I just hope you have some time to research and agree) or ignore or whatever. I will just assume agreed, disagreed, or ignored it.
* You do **not** need to respond/comment unless you want it. I will not push you into applying toolchains, reopening the PR or whatever unless you explicitly ask so.
* I would love to hear the other reasons to dislike toolchain for the build though. At the end of the day, I might be wrong, and it might be I should stop using `buildJdk` toolchain. However, `buildJdk` toolchain is the best option I know to enforce the project uses the latest LTS JDK for the build.

---

I suspect the blogpost meant that people should stop "using Java 1.8 toolchain to target Java 1.8 bytecode". I see no direct evidence of that in the post though.

I completely agree people should stop using Java 1.8 toolchain; and I am sure it is just a wrong usage of the toolchain feature. People should use the latest toolchain available, and use `--release` to target the generated code. That is effectively what the blogpost suggests in the end of the article, however, I do not understand why the blogpost suggests stop using toolchains for the compilation.
It might be the article is just out of date, however, the publication date seems recent.

/cc @JakeWharton

Jake, sorry for mentioning you, however, I would appreciate if you could restructure the blogpost. At the end of the day, the proper toolchain usage enables having the latest JDK for the build no matter which JVM starts the build. I truly do not understand why the blogpost dislikes "toolchain for the compilation". Sure you don't need to answer this, yet I hope to hearing some more insights from you.

-----

Detailed analysis:

> 1) Bad docs
> Last week I released a new version of [Retrofit](https://github.com/square/retrofit) which uses a Java toolchain to target Java 8. Its use of toolchains was contributed a while ago, and I simply forgot to remove it. As a consequence, [its Javadoc](https://square.github.io/retrofit/2.x/retrofit/) was built using JDK 8 and is thus not searchable. Searchable Javadoc came in [JEP 225](https://openjdk.org/jeps/225) with JDK 9.

The only outcome of that paragraph is **Retrofit** used an outdated toolchain for the build, and it is a completely opposite usage from what I suggest.
I suggest using **the latest** toolchain available.

In other words, "bad docs" issue does not exist.

> 2) Resource ignorance
> Without a Java toolchain, a modern JDK (even an outdated patch release of one) will honor resource limits and perform much better in containerized environments.

It is basically the same issue. It looks like the blogpost blames **old** JVM for not using resources properly, while the **build** toolchain should use **recent LTS** compiler version.
One again, the blogpost blames a non-existent issue.

> 3) Compiler bugs
> All software has bugs, and sometimes the JVM, the Java compiler, or both have bugs. When you are using a 10-year-old version of the JVM and Java compiler

Once again, the blogpost blames **old compilers**, while toolchain I suggest use the **latest** compiler version, and I do that especially to avoid old compiler bugs.

> 4) Worse performance
> Oracle and other large JVM shops devote lots of person-hours to making the JVM faster

Once again, the blogpost blames **old JVM** for being slow.
Well, who on Earth suggests using an **old** JVM toolchain for the build purposes? I suggest the latest LTS, so it should have the greatest performance.

> 5) Not needed for cross-compilation
>In JDK 9, however, this all changed. The compiler now contains a record of all public APIs from every Java version going back to Java 8. It also allows specifying a single compiler flag, --release, which sets the source code language version, the target bytecode version, and the available runtime APIs to the specified release. There is **simply no value in compiling with an older JDK** to target an older JVM anymore.

Exactly. That is why I suggest using toolchain to use the latest compiler always. The `--release` came as a different PR.

>6) Wasted disk space
> All those JDKs needlessly take up space in your home directory. Each JDK is a few hundred MiB

That is a wrong claim again. I'm not sure why the blogpost says "all those JDKs". What I suggest is to use the latest LTS for the build purposes. I do not suggest use "all the possible toolchain versions for the build". If you use LTS for the build, you do not get those "needlessly piling up JDKs".

>7)Not the Gradle JVM
> Toolchains are only used for tasks that create a new JVM. That means compilation (of Java or Kotlin) and running unit tests. They do not control the JVM that is used for running the actual Gradle build or any of the plugins therein

Of course, it would be great if Gradle's toolchains could control Gradle Daemon JVM (and I would love to have such a feature), however, Gradle engineers decided to use a completely different feature to control [Gradle Daemon JVM](https://docs.gradle.org/current/userguide/gradle_daemon.html#configuring_the_daemon_jvm).
Sure you one can blame toolchains for not being universal, however, the possibility to enforce Gradle Daemon JVM is there if one really wants it.

>8) Not all bad
> I want to stress that toolchains are unequivocally not a good idea for compilation. They still have utility elsewhere, however.

Frankly, I do not understand why the blogpost says "toolchains are not a good idea for compilation".
They literally say "As a result, we **compile with the latest Java**, but test through the lowest-supported Java"
Hey, they do compile with the latest Java, and the toolchains enables to enforce it. Java toolchains for the build enables to ensure everybody would build with a known Java version right as the blogpost suggests. I truly do not understand why they say "toolchains are not a good idea for compilation".

Issue #938, opened 2024-08-13
Redirect the old Maven coordinate to the new one


Goooler:
> Hi. Maven has a relocation mechanism which can be used for repositories: https://maven.apache.org/guides/mini/guide-relocation.html . The problem is that Gradle, as far as I know, does not support such a mechanism, so I am not sure it can actually be used automatically for Gradle plugins. It also requires a tombstone release to the original destination

_Originally posted by @oleg-nenashev in https://github.com/renovatebot/renovate/discussions/30676#discussioncomment-10320920_

Goooler:
But what, I just realized that Shadow plugin was only published to Gradle Plugin Portal, so we can't relocate it from MavenCentral.

The only thing we can do from our side might be https://github.com/gradle/plugin-portal-requests/issues/233.

Goooler:
See https://github.com/gradle/plugin-portal-requests/issues/233#issuecomment-2295907778.

Goooler:
Filed a new issue to https://github.com/dependabot/dependabot-core/issues/12947.

PR #1492, opened 2025-06-30
Remove commons-io dependency to reduce the surface for CVEs

vlsi:
---

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -76,10 +76,18 @@ publishing.publications.withType<MavenPublication>().configureEach {
 }
 
 dependencies {
+  constraints {
+    compileOnly("commons-io:commons-io") {
+      because("We do not want adding commons-io to the dependencies to reduce the surface for transitive CVEs")
+      version {
+        rejectAll()
+      }
+    }
+  }
+
   api("org.apache.ant:ant:1.10.15") // Types from Ant are exposed in the public API.
   implementation("org.jdom:jdom2:2.0.6.1")
   implementation("org.ow2.asm:asm-commons:9.8")
-  implementation("commons-io:commons-io:2.19.0")
   implementation("org.codehaus.plexus:plexus-utils:4.0.2")
   implementation("org.codehaus.plexus:plexus-xml:4.1.0")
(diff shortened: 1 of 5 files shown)


Goooler:
I wanted to do so, and I'd be happy to accept removing `commons-io`, but this is targeting 8.x; I believe we don't pay much focus on maintaining Shadow 8.

vlsi:
We use shadow 8.1.1 in https://github.com/pgjdbc/pgjdbc, and it results in having commons-io:commons-oi:2.11.0 on the build classpath which triggers  CVE 8.7 (CVE-2024-47554).
Sure we can try adding build-time constraint on commons-io in pgjdbc build scripts, however, I thought it would be better to fix the upstream.

vlsi:
As far as I can understand, [9.0.0 is in beta stage](https://central.sonatype.com/artifact/com.gradleup.shadow/shadow-gradle-plugin/versions), and it is too early to adopt, so it would be great if you could maintain 8.x for a while.
It is a sad case to have when 8.x has CVEs while 9.x is in beta.

Goooler:
`8.3.7` depends on `commons-io:2.19.0`, which is patched.

https://central.sonatype.com/artifact/com.gradleup.shadow/shadow-gradle-plugin/8.3.7/dependencies

vlsi:
Apparently, we use `com.github.johnrengelman.shadow`, and the coordinates changed to `com.gradleup.shadow` which defeated Renovate auto-update.

It would be nice if you could release a relocation pom for it (`com.github.johnrengelman.shadow:com.github.johnrengelman.shadow.gradle.plugin:...` -> `com.gradleup.shadow:com.gradleup.shadow.gradle.plugin:...`)

See https://github.com/renovatebot/renovate/issues/5667

An alternative/additional option could be filing a PR adding the replacement via https://github.com/renovatebot/renovate/blob/8d1202c71a3908b1defcd5bc5f8747d96eef732a/lib/data/replacements.json

Goooler:
I filed it, 

- https://github.com/renovatebot/renovate/discussions/30676
- #938

Issue #1483, opened 2025-06-24
Using PropertiesFileTransformer broken with version 8.3.7


cmuchinsky:
### Expected and Results

After upgrading from com.gradleup.shadow 8.3.6 to 8.3.7, my builds started failing with a `java.lang.StackOverflowError` exception:

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':xxx:shadowJar'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:130)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:293)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:128)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:116)
        at org.gradle.api.internal.tasks.execution.ProblemsTaskPathTrackingTaskExecuter.execute(ProblemsTaskPathTrackingTaskExecuter.java:41)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:74)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
[...]

PR #1493, opened 2025-07-01
Test `mergePropertiesWithDifferentStrategies` for `PropertiesFileTransformer`

Goooler:
Refs 

- #1483
- #1485

---

### DIFF
--- a/src/test/groovy/com/github/jengelman/gradle/plugins/shadow/TransformerSpec.groovy
+++ b/src/test/groovy/com/github/jengelman/gradle/plugins/shadow/TransformerSpec.groovy
@@ -2,6 +2,7 @@ package com.github.jengelman.gradle.plugins.shadow
 
 import com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer
 import com.github.jengelman.gradle.plugins.shadow.transformers.GroovyExtensionModuleTransformer
+import com.github.jengelman.gradle.plugins.shadow.transformers.PropertiesFileTransformer
 import com.github.jengelman.gradle.plugins.shadow.transformers.ServiceFileTransformer
 import com.github.jengelman.gradle.plugins.shadow.transformers.XmlAppendingTransformer
 import com.github.jengelman.gradle.plugins.shadow.util.PluginSpecification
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.groovy
```diff
@@ -136,7 +136,9 @@ class PropertiesFileTransformer implements Transformer, Serializable {
     String charset = 'ISO_8859_1'
 
     @Internal
-    Closure<String> keyTransformer = IDENTITY
+    Closure<String> keyTransformer = new Closure<String>("") {
+        String doCall(Object arguments) { arguments }
+    }
```
Goooler:
```java
    public static final Closure IDENTITY = new Closure<Object>((Object)null) {
        private static final long serialVersionUID = 730973623329943963L;

        public Object doCall(Object args) {
            return args;
        }
    };
```

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.groovy
```diff
@@ -136,7 +136,9 @@ class PropertiesFileTransformer implements Transformer {
     String charset = 'ISO_8859_1'
 
     @Internal
-    Closure<String> keyTransformer = IDENTITY
+    Closure<String> keyTransformer = new Closure<String>("") {
+        String doCall(Object arguments) { arguments }
```
Goooler:
```suggestion
        String doCall(Object arguments) { arguments } // We can't use Closure#IDENTITY instead, for being compatible with Groovy 3 and 4.
```

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.groovy
```diff
@@ -136,7 +136,9 @@ class PropertiesFileTransformer implements Transformer {
     String charset = 'ISO_8859_1'
 
     @Internal
-    Closure<String> keyTransformer = IDENTITY
+    Closure<String> keyTransformer = new Closure<String>("") {
+        String doCall(Object arguments) { arguments } // We can't use Closure#IDENTITY instead, for being compatible with Groovy 3 and 4.
```
Goooler:
```suggestion
        String doCall(Object arguments) { arguments } // We can't use Closure#IDENTITY instead, as it is not compatible with Groovy 3 and 4.
```

Issue #1416, opened 2025-04-28
`shadowJar` task does not preserve JVM version attribute


Jonathing:
### Expected and Results

When a Jar is created with the `jar` task, the JVM version is saved as an attribute (`org.gradle.jvm.version`) for consumption in the Gradle Module metadata. The `shadowJar` task does not preserve this attribute, leading to strange resolutions on consumer projects if their project version is lower than the standard `runtimeElements` variant. The `shadowRuntimeElements` variant does not have this attribute, which means there is no way to check if the JVM version is higher than available on that variant, so it immediately attempts to select it.

For reference, here a diff if what the shadow variant is currently missing:
```diff
 "attributes": {
   "org.gradle.category": "library",
   "org.gradle.dependency.bundling": "shadowed",
-  "org.gradle.jvm.version": 17,
   "org.gradle.libraryelements": "jar",
   "org.gradle.usage": "java-runtime"
 },
```

### Related environent and versions

Gradle 8.14
Shadow 8.3.6

### Reproduction steps

1. Depend on a library which has a standard variant and a shadow variant.
    - The shadow variant must not have the target JVM version attribute.
    - The standard variant musth have the target JVM version attribute with a value higher than your project's toolchain language version.
2. Upon dependency resolution, Gradle will select the shadow variant as it does not have the target JVM version attribute.

### Anything else?

Here's how I work around this when publishing fat-jar artifacts (Groovy DSL):
```gradle
import org.gradle.api.attributes.java.TargetJvmVersion
import com.github.jengelman.gradle.plugins.shadow.ShadowJavaPlugin

configurations {
    named(ShadowJavaPlugin.SHADOW_RUNTIME_ELEMENTS_CONFIGURATION_NAME) {
        attributes {
            attribute TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE, java.toolchain.languageVersion.get().asInt()
        }
    }
}
```

Goooler:
This should be fixed in #1199.

Jonathing:
Ah. 9.0 beta then? Thank you!

Goooler:
Yes

PR #1502, opened 2025-07-06
Consider release option for target JVM attribute

### LINKED ISSUE #1501: Target JVM version is incorrectly inferred
### Expected and Results

#1199 introduced the missing `"org.gradle.jvm.version"` attribute for the module metadata of the shadowed artifacts. The version is however not resolved to the correct JVM version in all cases.

### Related environment and versions

Shadow: 9.0.0-rc1

### Reproduction steps

Consider the following example:
```kotlin
java {
	toolchain {
		languageVersion.set(JavaLanguageVersion.of(21))
	}
}

tasks.compileJava {
	options.release = 8
}
```
(See [this blog post](https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/) for why you'd want to do this)

This produces a jar compatible with Java 8, but when also applying the shadow plugin it claims it was compiled against Java 21.

### Anything else?

The implementation currently only looks at `targetCompatibility` https://github.com/GradleUp/shadow/blob/b224199bfd97b937b1e965d43c71193bac9da360/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt#L78-L82

The corresponding implementation in Gradle seems to take a few more cases into consideration:
https://github.com/gradle/gradle/blob/b06b7de4572c102a02ed831ae1f37e2da8255c35/platforms/jvm/plugins-java-base/src/main/java/org/gradle/api/plugins/jvm/internal/DefaultJvmLanguageUtilities.java#L64-L114

DreierF:
Fixes #1501

---

### DIFF
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/PublishingTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/PublishingTest.kt
@@ -87,6 +87,7 @@ class PublishingTest : BasePluginTest() {
     }.toTypedArray()
     val targetJvmAttr17 = TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE.name to "17"
     val targetJvmAttr11 = TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE.name to "11"
+    val targetJvmAttr8 = TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE.name to "8"
 
     projectScriptPath.appendText(
       """
@@ -118,6 +119,23 @@ class PublishingTest : BasePluginTest() {
     publish()
     // sourceCompatibility doesn't affect the target JVM version.
     assertions(attrsWithoutTargetJvm + targetJvmAttr11)
+
+    projectScriptPath.appendText(
+      """
+        java {
+          toolchain {
+            languageVersion.set(JavaLanguageVersion.of(21))
+          }
+        }
+
+        tasks.compileJava {
+          options.release = 8
+        }
+      """.trimIndent() + System.lineSeparator(),
+    )
+    publish()
+    // release flag is considered.
+    assertions(attrsWithoutTargetJvm + targetJvmAttr8)
   }
 
   @Test
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
@@ -75,9 +75,13 @@ public abstract class ShadowJavaPlugin @Inject constructor(
           objects.named(LibraryElements::class.java, LibraryElements.JAR),
         )
         attr.attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling::class.java, Bundling.SHADOWED))
-        val targetJvmVersion = provider {
-          javaPluginExtension.targetCompatibility.majorVersion.toInt()
-        }
+        val targetJvmVersion =
+          configurations.named(JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME)
+            .map {
+              it.attributes.getAttribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE)
+                ?: javaPluginExtension.targetCompatibility.majorVersion.toInt()
+            }
+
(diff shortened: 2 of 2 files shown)


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/PublishingTest.kt
```diff
@@ -118,6 +119,23 @@ class PublishingTest : BasePluginTest() {
...
+      """
+        java {
+          toolchain {
+            languageVersion.set(JavaLanguageVersion.of(21))
+          }
+        }
+
+        tasks.compileJava {
+          options.release = 8
+        }
```
Goooler:
```suggestion
        tasks.named('compileJava') {
          options.release = 8
        }
```

docs/changes/README.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased](https://github.com/GradleUp/shadow/compare/9.0.0-rc1...HEAD) - 2025-xx-xx
 
+**Fixed**
+- JVM version in module metadata for published jar was wrong when target was set via `options.release` ([#1502](https://github.com/GradleUp/shadow/pull/1502))
```
```suggestion
- Honor `options.release` for target JVM attribute. ([#1502](https://github.com/GradleUp/shadow/pull/1502))
```

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
```diff
@@ -75,9 +75,13 @@ public abstract class ShadowJavaPlugin @Inject constructor(
...
         attr.attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling::class.java, Bundling.SHADOWED))
-        val targetJvmVersion = provider {
-          javaPluginExtension.targetCompatibility.majorVersion.toInt()
-        }
+        val targetJvmVersion =
+          configurations.named(JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME)
+            .map {
+              it.attributes.getAttribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE)
+                ?: javaPluginExtension.targetCompatibility.majorVersion.toInt()
+            }
```
Goooler:
```suggestion
        val targetJvmVersion = configurations.named(JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME)
          .map { compileClasspath ->
            compileClasspath.attributes.getAttribute(TargetJvmVersion.TARGET_JVM_VERSION_ATTRIBUTE)
              ?: javaPluginExtension.targetCompatibility.majorVersion.toInt()
          }
```

Goooler (approved):

src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/PublishingTest.kt
```diff
@@ -118,6 +119,17 @@ class PublishingTest : BasePluginTest() {
...
+
+    projectScriptPath.appendText(
+      """
+        tasks.named('compileJava') {
+          options.release = 8
+        }
+      """.trimIndent() + System.lineSeparator(),
+    )
+    publish()
+    // release flag is considered.
```
Goooler:
```suggestion
    // options.release flag is honored.
```

martinbonnin:
Been looking at this in the context of https://github.com/GradleUp/shadow/issues/1540 and I don't think there is a good value to put in `org.gradle.jvm.version`? 

If we really wanted to put something here, the best value is probably the minimum of all shadowed jars:

```
min(
  this-project jvmTarget, 
  lib1 org.gradle.jvm.version,
  lib2 org.gradle.jvm.version,
  ... 
)
```

Even that feels a bit wrong. If someone is using the fat jar as a compile dependency, only the public API surface needs to be say Java 8. Whether the shadowed libs are Java 11 or 17 doesn't really matter (see also https://github.com/gradle/gradle/issues/33964).

But with the current Gradle behaviour, I guess we don't really have a choice if we also want to fix https://github.com/GradleUp/shadow/issues/1416

Goooler:
I'm afraid we should remove `org.gradle.jvm.version`, this is the least disruptive way to maintain compatibility in cases like #1540.

martinbonnin:
> we should remove org.gradle.jvm.version

👍 This is fine by me. The more I think about this, the more I think the JVM target shouldn't be part of dependency resolution anyways

PR #1505, opened 2025-07-10
Restore Develocity Build Scan integration

### LINKED ISSUE #907: Plugin is incompatible with isolated projects feature

gabrielfeo:
Restore the Develocity Build Scan integration, which adds custom values to Build Scans about the execution of Shadow tasks. It had been first disabled (#993) to resolve #907 for consumers enabling [Isolated Projects][1] (IP), then removed altogether (#1014). 

- Code is added similar to what was removed #1014, but with changes to migrate it to Kotlin (it was Groovy)
- The code is made IP-compatible by using the `buildScan` extension that's available in subprojects (cd9c60071fa5af35a3bd82cd695c121c7ae34322), instead of the root project's extension. That was added in [Develocity Gradle plugin 4.0][2].
- A `compileOnly` dependency is added for `ShadowPlugin` to reference the `buildScan` extension class (that wasn't necessary in Groovy)
- Support for `com.gradle.enterprise` which was present in the original code is omitted. There's no version of that plugin supporting the `buildScan` extension in subprojects, thus adding custom values would have to take place in the root project (not IP-compatible). That plugin is replaced by `com.gradle.develocity` and no longer maintained.

Currently, the custom values added by this PR are an example, since `ShadowStats` was also removed recently (#1264).

### Tests

A functional test around IP compatibility is added to demonstrate that the fix works and to prevent future regressions.

- [Build Scan showing the test failing before the fix][3], reproducing the same error reported in the original issue (#907)
- [Build Scan showing the test passing after the fix][4]

[1]: https://docs.gradle.org/current/userguide/isolated_projects.html
[2]: https://docs.gradle.com/develocity/gradle-plugin/current/#4_0
[3]: https://scans.gradle.com/s/r3yoxewbrtx5g/tests/task/:functionalTest/details/com.github.jengelman.gradle.plugins.shadow.IsolatedProjectsCompatibilityTest/applyToProjectWithIsolatedProjectsEnabled()?top-execution=1
[4]: https://scans.gradle.com/s/y2dxeizmtcobe/tests/task/:functionalTest/details/com.github.jengelman.gradle.plugins.shadow.IsolatedProjectsCompatibilityTest/applyToProjectWithIsolatedProjectsEnabled()?top-execution=1

---

### DIFF
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
@@ -41,6 +41,7 @@ import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.plugins.JavaPlugin.API_CONFIGURATION_NAME
 import org.gradle.api.plugins.JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME
 import org.gradle.api.tasks.bundling.Jar
+import org.gradle.language.base.plugins.LifecycleBasePlugin
 import org.gradle.testfixtures.ProjectBuilder
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.params.ParameterizedTest
@@ -64,6 +65,8 @@ class JavaPluginTest : BasePluginTest() {
     project.plugins.apply(JavaPlugin::class.java)
     val shadowTask = project.tasks.getByName(SHADOW_JAR_TASK_NAME) as ShadowJar
     val shadowConfig = project.configurations.getByName(ShadowBasePlugin.CONFIGURATION_NAME)
+    val assembleTask = project.tasks.getByName(LifecycleBasePlugin.ASSEMBLE_TASK_NAME)
+    assertThat(assembleTask.dependsOn).contains(shadowTask)
 
     // Check extended properties.
     with(shadowTask as Jar) {
@@ -93,13 +96,6 @@ class JavaPluginTest : BasePluginTest() {
     }
 
     assertThat(shadowConfig.artifacts.files).contains(shadowTask.archiveFile.get().asFile)
-  }
-
-  @Test
-  fun incompatibleWithLowerMinGradleVersion() {
-    runWithFailure(shadowJarTask) {
-      it.withGradleVersion("8.2")
-    }
   }
 
   @Test
@@ -600,6 +596,26 @@ class JavaPluginTest : BasePluginTest() {
     val result = runWithFailure(shadowJarTask)
 
     assertThat(result.output).containsMatch("Cannot expand ZIP '.*bad\\.jar'".toRegex())
+  }
+
+  @Test
+  fun failBuildIfProcessingAar() {
+    val fooAarPath = path("foo.aar")
+
+    projectScriptPath.appendText(
+      """
+        dependencies {
+          ${implementationFiles(fooAarPath)}
+        }
+      """.trimIndent(),
+    )
+
+    val result = runWithFailure(shadowJarTask)
+
+    assertThat(result.output).contains(
+      "Shadowing AAR file is not supported.",
+      "Please exclude dependency artifact:",
+    )
   }
 
   @Test
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPlugin.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPlugin.kt
@@ -24,6 +24,12 @@ public abstract class ShadowPlugin : Plugin<Project> {
     withId(KOTLIN_MULTIPLATFORM_PLUGIN_ID) {
       apply(ShadowKmpPlugin::class.java)
     }
+    withId("com.android.base") {
+      error(
+        "Shadow does not support using with AGP, you may need Android Fused Library plugin instead. " +
+          "See https://developer.android.com/build/publish-library/fused-library",
+      )
+    }
     project.configureBuildScan()
 
     // Apply the legacy plugin last.
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -13,7 +13,14 @@
       "matchPackageNames": [
         "org.vafer:jdependency",
         "org.ow2.asm:asm-commons"
-      ]
-    }
+      ],
+    },
+    {
+      // AGP version should match the min Gradle version used in tests.
+      "enabled": false,
+      "matchPackageNames": [
+        'com.android.tools.build:gradle',
+      ],
+    },
   ]
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,7 +4,7 @@ import org.gradle.api.plugins.JavaPlugin.API_ELEMENTS_CONFIGURATION_NAME
 import org.gradle.api.plugins.JavaPlugin.JAVADOC_ELEMENTS_CONFIGURATION_NAME
 import org.gradle.api.plugins.JavaPlugin.RUNTIME_ELEMENTS_CONFIGURATION_NAME
 import org.gradle.api.plugins.JavaPlugin.SOURCES_ELEMENTS_CONFIGURATION_NAME
-import org.gradle.kotlin.dsl.kotlin
+import org.jetbrains.kotlin.daemon.common.OSKind
 import org.jetbrains.kotlin.gradle.dsl.JvmDefaultMode
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
@@ -118,6 +118,7 @@ dependencies {
   implementation(libs.plexus.utils)
   implementation(libs.plexus.xml)
 
+  testPluginClasspath(libs.agp)
   testPluginClasspath(libs.foojayResolver)
   testPluginClasspath(libs.develocity)
   testPluginClasspath(libs.kotlin.kmp)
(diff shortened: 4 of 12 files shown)


gabrielfeo:
Hi there 👋🏻 

This PR demonstrates that Develocity/Build Scan integration could be restored, but it's not clear to me if you'd like to restore `ShadowStats` as well or add different custom values. We could add actual custom values and also create more tests.

Goooler:
Restoring `ShadowStats` is hard, and it hasn't been tested. I suspect there are numerous issues regarding its accuracy.

Actually, I don't have much interest in adding `ShadowStats` or Develocity integration back.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPlugin.kt
```diff
@@ -29,5 +31,36 @@ public abstract class ShadowPlugin : Plugin<Project> {
     // If the user applies shadow before those plugins. However, this is fine, because this was also
     // the behavior with the old plugin when applying in that order.
     apply(LegacyShadowPlugin::class.java)
+
+    withId("com.gradle.develocity") {
```
wolfs: ❌ I don't think this will work since the Develocity plugin isn't applied to subprojects. The test doesn't detect this, because it doesn't have a sub-project. I suggest checking whether an extension with the name `develocity` is present at the project level.

wolfs (changes requested):

src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
```diff
@@ -779,6 +781,36 @@ class JavaPluginTest : BasePluginTest() {
...
+    settingsScriptPath.writeText(
+      """
+        plugins {
+          id 'com.gradle.develocity'
+        }
+        develocity {
+          buildScan {
+            termsOfUseUrl = 'https://gradle.com/terms-of-service'
+            termsOfUseAgree = 'yes'
+            publishing.onlyIf { true }
```
wolfs: ❌ I don't think it is a good idea for the test to publish a build scan to scans.gradle.com. Should we dump the scan file instead.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPlugin.kt
gabrielfeo: Thank you, @wolfs. You're correct. I've changed the code to your suggestion in d7025200f941f113ea4e9e5c6ac7c1be33869ac5.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/BuildScanIntegration.kt
```diff
@@ -0,0 +1,18 @@
+package com.github.jengelman.gradle.plugins.shadow.internal
+
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+import com.gradle.develocity.agent.gradle.DevelocityConfiguration
+import org.gradle.api.Project
+
+public fun Project.addBuildScanCustomValues() {
+  extensions.configure<DevelocityConfiguration>("develocity") {
```
wolfs: ❓ Why do you obtain the extension again, instead of casing the one you got via `project.extensions.findByName("develocity")`?

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/BuildScanIntegration.kt
```diff
@@ -0,0 +1,18 @@
...
+    val buildScan = it.buildScan
+    val shadowTasks = tasks.withType(ShadowJar::class.java)
+    shadowTasks.configureEach { task ->
+      buildScan.buildFinished {
```
wolfs: This is now happening when the shadow task is being configured. I suppose you want to add it only when the shadow task has been executed. Was this like this before?

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/BuildScanIntegration.kt
gabrielfeo: It seemed a (strong typing) code smell to take an `Any` argument here and cast it. On the other hand, I wouldn't be able to take a `DevelocityConfiguration`-typed argument here because that class is in a `compileOnly` dependency, meaning I couldn't reference it as eagerly as in the plugin class if it's not present at runtime after all (783860bf228a6b110c9d2bc9a507e5c970c7423c).

To still address your comment though, I removed the extension lookup that was in the plugin and kept this one (826bb4103f678f09654f8a3e607075c065165bef).

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/BuildScanIntegration.kt
gabrielfeo: The original code was slightly different. It added a single `buildFinished` callback and iterated on each tasks [inside](https://github.com/GradleUp/shadow/pull/1014/files#diff-4f688def42f06583701e7481cc7f29c906b6c9cf0808dac578d46d6300781ec9L46) that callback, but that requires serializing `Project` for configuration cache as part of that callback, which isn't supported. It also achieved what you propose by using `didWork`, which I also added (34a18ac6385d03dd71efe83a0327a799ebd3b71e), but it won't work here if I iterate on tasks at configuration time (and keep CC compatibility)

src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
gabrielfeo: While the test doesn't currently depend on/assert about scan publishing having succeeded, I think that's a good point. I changed it to dump the scan file in d639b8a076d4b62d37c48ed2f6bfda184b212d34.

Goooler: Removing `develocity` block in e829b8a02aae585024b4f776c6dacfa0926ae290.

gabrielfeo:
> Restoring ShadowStats is hard, and it hasn't been tested. I suspect there are numerous issues regarding its accuracy.
> 
> Actually, I don't have much interest in adding ShadowStats or Develocity integration back.

@Goooler, aside from `ShadowStats`, is there any metadata or measurements you think would be useful to surface in a Build Scan that we could help implement?

The benefit for users would be, when viewing a single Build Scan, this metadata giving insights into work shadow did without having to add to the console log. Also, when using a Develocity server to process several Build Scans at scale, tracking certain measurements can help find bottlenecks or unwanted usages across several projects (e.g. dependencies and configurations as the integration logged before), fix them and track the impact.

Goooler:
I believe this is a secondary feature with low priority. If you really want to reintroduce it, please include it in the follow-ups and ensure it is thoroughly tested.

Goooler:
@wolfs Any comments?

Goooler:
CC @wolfs

Issue #484, opened 2019-04-25
Shadow is not working with Kotlin Multiplatform plugin


Danilo-Araujo-Silva:
Please check the [User Guide](http://imperceptiblethoughts.com/shadow) before submitting "how do I do 'x'?" questions!

### Shadow Version
5.0.0

### Gradle Version
5.3.1

### Expected Behavior
The `shadowJar` task should appear and work.

### Actual Behavior
Only the `knows` task appears and the shadow task don't appear and don't work.

### Gradle Build Script(s)
```
import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
...
plugins {
    kotlin("multiplatform")

    id("com.github.johnrengelman.shadow") version "5.0.0"
}
...
tasks.withType<ShadowJar> {
    baseName = "app"
    classifier = ""
    version = ""
}

//// This works but do not really creat the fatJar, java -jar myFatJar.jar doesn't run and inside of this jar we have just a MANIFEST.MF with a single line of information.
//tasks.create("shadowJar", ShadowJar::class.java) {
//	baseName = "my_project"
//	classifier = ""
//	version = ""
//}

```

### Content of Shadow JAR (`jar tf <jar file>` - post link to GIST if too long)
None.

Danilo-Araujo-Silva:
I was able to have something working with:

```
import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
...
plugins {
	kotlin("multiplatform")

	id("com.github.johnrengelman.shadow") version "5.0.0"
	id("java")
}
...
tasks.withType<ShadowJar> {
    manifest {
        attributes.put("Main-Class", "com.example.ExampleKt")
    }

    val target = kotlin.targets.jvm("myTarget")
    from(target.compilations["main"].output)
    val runtimeClasspath = target.compilations["main"].compileDependencyFiles as Configuration
    configurations = mutableListOf(runtimeClasspath)
}
```

But I need to be checking the other options available.

Ingwersaft:
Your workaround didn't work for me, I had to modify it a little bit, maybe it helps someone :)

```
import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar

plugins {
    kotlin("multiplatform") version "1.3.50"
    id("com.github.johnrengelman.shadow") version "5.1.0"
}
val ktor_version: String by project
val kotlin_version: String by project
val logback_version: String by project

repositories {
    mavenCentral()
    mavenLocal()
[...]

PR #1573, opened 2025-08-03
Don't register `ShadowJar` tasks for multiple KMP JVM targets

Goooler:
Refs #484.


Goooler:
```
> Configure project :
e: ❌ `jvm()` Kotlin Target Already Declared
Declaring multiple Kotlin Targets of the same type is not supported.
Solution: Please remove the duplicate target declaration.
See https://kotl.in/declaring-multiple-targets for more details.
```

PR #1576, opened 2025-08-03
Normalize functional test names

Goooler:


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/JavaPluginTest.kt
```diff
@@ -720,7 +720,7 @@ class JavaPluginTest : BasePluginTest() {
   }
 
   @Test
-  fun addExtraFilesIntoShadowJar() {
+  fun canAddExtraFilesIntoShadowJar() {
```
Goooler: So strange! After adding the `can` prefix back, the test passed....

PR #1579, opened 2025-08-04
Use `BufferedOutputStream` when writing the Zip file to improve performance

stuartwdouglas:
When not using STORED entries use a BufferedOutputStream to avoid lots of small writes to the file system.

Testing this with a 300mb jar build I see the total build time going from 40s to 30s.

Note that it is not possible to do this with STORED entries as the implementation requires a RandomAccessFile to update the CRC after write.

---

### DIFF
--- a/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultZipCompressor.groovy
+++ b/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultZipCompressor.groovy
@@ -31,8 +31,8 @@ class DefaultZipCompressor implements ZipCompressor {
     ZipOutputStream createArchiveOutputStream(File destination) {
         try {
             ZipOutputStream zipOutputStream = entryCompressionMethod == ZipOutputStream.STORED ?
-                // It is not possible to do this with STORED entries as the implementation requires a RandomAccessFile to update the CRC after write.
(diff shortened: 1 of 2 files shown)


Goooler (approved):

src/main/groovy/com/github/jengelman/gradle/plugins/shadow/internal/DefaultZipCompressor.groovy
```diff
@@ -30,7 +30,10 @@ class DefaultZipCompressor implements ZipCompressor {
     @Override
     ZipOutputStream createArchiveOutputStream(File destination) {
         try {
-            ZipOutputStream zipOutputStream = new ZipOutputStream(destination)
+            ZipOutputStream zipOutputStream = entryCompressionMethod == ZipOutputStream.STORED ?
+                // It is not possible to do this with STORED entries as the implementation requires a RandomAccessFile to update the CRC after write.
+                new ZipOutputStream(destination) :
+                new ZipOutputStream(new BufferedOutputStream(new FileOutputStream(destination)))
```
Goooler:
```suggestion
                new ZipOutputStream(destination) :
                // It is not possible to do this with STORED entries as the implementation requires a RandomAccessFile to update the CRC after write.
                new ZipOutputStream(new BufferedOutputStream(new FileOutputStream(destination)))
```

PR #1580, opened 2025-08-04
Use `BufferedOutputStream` when writing the Zip file to improve performance

stuartwdouglas:
When not using STORED entries use a BufferedOutputStream to avoid lots of small writes to the file system.

Testing this with a 300mb jar build I see the total build time going from 40s to 30s.

Note that it is not possible to do this with STORED entries as the implementation requires a RandomAccessFile to update the CRC after write.

---


stuartwdouglas:
I didn't add a changelog entry for this as if the 8.x PR goes in a well then there are no changes compared to the 8.x branch.

Goooler (approved):

PR #1601, opened 2025-08-07
Improve the error message for empty `mainClassName`

Goooler:
---


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowApplicationPlugin.kt
```diff
@@ -129,11 +129,14 @@ public abstract class ShadowApplicationPlugin : Plugin<Project> {
   protected open fun Project.configureShadowJarMainClass() {
     val mainClassName = applicationExtension.mainClass
     tasks.shadowJar.configure { task ->
-      task.inputs.property("mainClassName", mainClassName)
```
Goooler, line 132: This will result in the `shadowJar` task not being re-run after `applicationExtension.mainClass` changed.

Goooler: #1603

PR #1642, opened 2025-08-12
Share `AppendableMavenRepository` for tests

Goooler:


Goooler:
We shouldn't share it...

Goooler:
Things are addressed into #1646.

PR #1646, opened 2025-08-13
Support publishing all modules at once

Goooler:
Follow up #1639.


Goooler:
This improved the test performance!

PR #1662, opened 2025-08-16
Allow opting out of `shadowRuntimeElements` variant

### LINKED ISSUE #651: Unable to publish only unshadowed artifacts

Goooler:
Closes #651.

---

### DIFF
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
@@ -92,7 +92,7 @@ public abstract class ShadowJavaPlugin @Inject constructor(
       variant.mapToMavenScope("runtime")
     }
     afterEvaluate {
-      if (!shadow.addOptionalJavaVariant.get()) return@afterEvaluate
+      if (!shadow.addShadowVariantIntoJavaComponent.get()) return@afterEvaluate
       components.named("java", AdhocComponentWithVariants::class.java) {
         it.addVariantsFromConfiguration(shadowRuntimeElements) { variant ->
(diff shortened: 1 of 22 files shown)


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
```diff
@@ -86,16 +86,19 @@ public abstract class ShadowJavaPlugin @Inject constructor(
...
     shadowComponent.addVariantsFromConfiguration(shadowRuntimeElements) { variant ->
       variant.mapToMavenScope("runtime")
     }
+    afterEvaluate {
+      if (!shadow.addShadowVariantIntoJavaComponent.get()) return@afterEvaluate
+      components.named("java", AdhocComponentWithVariants::class.java) {
+        it.addVariantsFromConfiguration(shadowRuntimeElements) { variant ->
+          variant.mapToOptional()
+        }
+      }
```
Goooler, line 100: It should be skipped by `skip`.

Goooler: #2056

PR #1682, opened 2025-08-21
Document more about `eachFile`, `filesMatching`, and `filesNotMatching`

### LINKED ISSUE #1677: Feature Request: Fine-Grained Content Filtering for Dependencies

We are currently facing an issue where our project (JAR A) depends on a shaded JAR (JAR B). This shaded JAR B includes an older, vulnerable version of a library (Package C), including its META-INF files.

Our project A also brings in a newer, non-vulnerable version of Package C as a transitive dependency. However, during the final shading process for JAR A, the older META-INF files from Package C, which are bundled inside the shaded JAR B, are being included. This is causing our vulnerability scanner to flag the final artifact as having a known vulnerability, even though the newer classes of Package C are correctly used at runtime.

We need a mechanism to precisely exclude the problematic META-INF directory of the old Package C that resides within the shaded JAR B.

---

We would like the Gradle Shadow plugin to implement a feature for fine-grained content filtering on a per-dependency basis. This functionality is already available in the Apache Maven Shade Plugin (https://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html) and is extremely powerful for resolving these types of conflicts. 

Here is an example from their documentation of how they achieve this:

```xml
<filters>
  <filter>
    <artifact>junit:junit</artifact>
    <includes>
      <include>junit/framework/**</include>
    </includes>
    <excludes>
      <exclude>org/junit/runners/**</exclude>
    </excludes>
  </filter>
  <filter>
    <artifact>*:*</artifact>
    <excludes>
      <exclude>META-INF/*.SF</exclude>
      <exclude>META-INF/*.DSA</exclude>
      <exclude>META-INF/*.RSA</exclude>
    </excludes>
  </filter>
</filters>
```

For our specific use case in gradle, it might look something like this:

```groovy
// Hypothetical DSL
shadowJar {
  dependencies {
    // ... other configurations
    include(dependency('group:shaded-jar-B')) {
      exclude('META-INF/maven/vulnerable/package-c/**')
    }
}
```

This would instruct the Shadow plugin to remove all files matching the exclude pattern, but only from the contents of the shaded-jar-B dependency before merging them into the final uber JAR.

Thanks for your consideration!

### Upstream changes

n/a

Yes if the proposal is sound.

Goooler:
Closes #1677.

### DIFF
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/ServiceFileTransformerTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/ServiceFileTransformerTest.kt
@@ -262,38 +262,20 @@ class ServiceFileTransformerTest : BaseTransformerTest() {
     }
   }
 
-  @Test
-  fun strategyExcludeCanBeOverriddenByEachFile() {
-    writeDuplicatesStrategy(DuplicatesStrategy.EXCLUDE)
+  @ParameterizedTest
+  @MethodSource("eachFileStrategyProvider")
+  fun strategyCanBeOverriddenByEachFile(
+    default: DuplicatesStrategy,
+    override: DuplicatesStrategy,
+    matchPath: String,
+  ) {
+    writeDuplicatesStrategy(default)
     projectScript.appendText(
       """
         $shadowJarTask {
           eachFile {
-            if (path == '$ENTRY_SERVICES_SHADE') {
-              duplicatesStrategy = DuplicatesStrategy.INCLUDE
-            }
-          }
-        }
-      """.trimIndent(),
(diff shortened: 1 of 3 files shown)


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/ServiceFileTransformerTest.kt
```diff
@@ -276,24 +309,24 @@ class ServiceFileTransformerTest : BaseTransformerTest() {
     )
   }
 
-  private companion object {
+  companion object {
```
Goooler:
```suggestion
  private companion object {
```

PR #1694, opened 2025-08-27
Do not write modified class files for no-op relocations

msawyer-palantir:
If we don't need to make any changes to the class, we should use the original bytes instead of letting ASM generate new ones. 
Some tools (e.g. https://plugins.gradle.org/plugin/com.palantir.baseline-class-uniqueness) use (hashes of) class file bytes to determine if there are classpath conflicts, and Shadow's current logic causes false positives.
But even in general, I believe it makes sense to change as little as we can.

A similar change was made to the maven shade plugin at https://github.com/apache/maven-shade-plugin/pull/95.

---

### DIFF
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/BasePluginTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/BasePluginTest.kt
@@ -58,6 +58,7 @@ abstract class BasePluginTest {
 
   val projectScript: Path get() = path("build.gradle")
   val settingsScript: Path get() = path("settings.gradle")
+  val outputJar: JarPath get() = jarPath("build/libs/my-1.0.jar")
   open val outputShadowedJar: JarPath get() = jarPath("build/libs/my-1.0-all.jar")
   val outputServerShadowedJar: JarPath get() = jarPath("server/build/libs/server-1.0-all.jar")
 
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/RelocationTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/RelocationTest.kt
@@ -10,8 +10,8 @@ import assertk.fail
 import com.github.jengelman.gradle.plugins.shadow.internal.mainClassAttributeKey
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowCopyAction.Companion.CONSTANT_TIME_FOR_ZIP_ENTRIES
 import com.github.jengelman.gradle.plugins.shadow.util.Issue
-import com.github.jengelman.gradle.plugins.shadow.util.JarPath
 import com.github.jengelman.gradle.plugins.shadow.util.containsOnly
+import com.github.jengelman.gradle.plugins.shadow.util.getBytes
 import com.github.jengelman.gradle.plugins.shadow.util.runProcess
 import java.net.URLClassLoader
 import kotlin.io.path.appendText
@@ -542,15 +542,15 @@ class RelocationTest : BasePluginTest() {
           implementation 'junit:junit:3.8.2'
         }
         $shadowJarTask {
-          relocate('junit', 'shadow.junit')
+          enableAutoRelocation = true
         }
       """.trimIndent(),
     )
 
     run(":jar", shadowJarPath)
 
-    val originalBytes = outputJar.getEntryBytes(mainClassEntry)
-    val relocatedBytes = outputShadowedJar.getEntryBytes(mainClassEntry)
+    val originalBytes = outputJar.use { it.getBytes(mainClassEntry) }
+    val relocatedBytes = outputShadowedJar.use { it.getBytes(mainClassEntry) }
     assertThat(relocatedBytes).isEqualTo(originalBytes)
   }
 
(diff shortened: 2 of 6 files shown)


src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/BasePluginTest.kt
```diff
@@ -58,6 +58,7 @@ abstract class BasePluginTest {
 
   val projectScript: Path get() = path("build.gradle")
   val settingsScript: Path get() = path("settings.gradle")
+  val outputThinJar: JarPath get() = jarPath("build/libs/my-1.0.jar")
```
Goooler:
```suggestion
  val outputJar: JarPath get() = jarPath("build/libs/my-1.0.jar")
```

src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/RelocationTest.kt
```diff
@@ -531,6 +533,29 @@ class RelocationTest : BasePluginTest() {
...
+      """.trimIndent(),
+    )
+
+    run(":jar")
+    val originalBytes = outputThinJar.getEntryBytes(mainClassEntry)
+
+    run(shadowJarPath)
+    val relocatedBytes = outputShadowedJar.getEntryBytes(mainClassEntry)
+
+    assertThat(relocatedBytes).isEqualTo(originalBytes)
```
Goooler:
```suggestion
    run(":jar", shadowJarPath)

    val originalBytes = outputThinJar.getEntryBytes(mainClassEntry)
    val relocatedBytes = outputShadowedJar.getEntryBytes(mainClassEntry)
    assertThat(relocatedBytes).isEqualTo(originalBytes)
```

src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/RelocationTest.kt
```diff
@@ -549,6 +574,11 @@ class RelocationTest : BasePluginTest() {
     }
   }
 
+  private fun JarPath.getEntryBytes(entryName: String): ByteArray = use { jar ->
```
Goooler: This could be placed in `JarPath.kt`.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.kt
```diff
@@ -201,6 +203,13 @@ public open class ShadowCopyAction(
         throw GradleException("Error in ASM processing class $path", t)
       }
 
+      val newClassBytes = if (hasModified) {
```
Goooler: Seems there is no need to add a `modifiedCallback`, all logic could be addressed before this. The `remapClass` could be something like:


```kt
    private fun FileCopyDetails.remapClass() = file.inputStream().use { inputStream ->
      // Temporarily remove the multi-release prefix.
      val multiReleasePrefix = "^META-INF/versions/\\d+/".toRegex().find(path)?.value.orEmpty()
      val newPath = path.replace(multiReleasePrefix, "")
      val mappedName = multiReleasePrefix + remapper.mapPath(newPath)
      val mappedPath = "$mappedName.class"

      val bytes = if (mappedPath == file.path) {
        // Write class bytes unmodified if possible.
        inputStream.readBytes()
      } else {
        // We don't pass the ClassReader here. This forces the ClassWriter to rebuild the constant pool.
        // Copying the original constant pool should be avoided because it would keep references
        // to the original class names. This is not a problem at runtime (because these entries in the
        // constant pool are never used), but confuses some tools such as Felix's maven-bundle-plugin
        // that use the constant pool to determine the dependencies of a class.
        val cw = ClassWriter(0)
        val cr = ClassReader(inputStream)
        val cv = ClassRemapper(cw, remapper)

        try {
          cr.accept(cv, ClassReader.EXPAND_FRAMES)
          cw.toByteArray()
        } catch (t: Throwable) {
          throw GradleException("Error in ASM processing class $path", t)
        }
      }

      try {
        val entry = zipEntry(mappedPath, preserveFileTimestamps, lastModified) {
          unixMode = UnixStat.FILE_FLAG or permissions.toUnixNumeric()
        }
        // Now we put it back on so the class file is written out with the right extension.
        zipOutStr.putNextEntry(entry)
        zipOutStr.write(bytes)
        zipOutStr.closeEntry()
      } catch (_: ZipException) {
        logger.warn("We have a duplicate $mappedName in source project")
      }
    }
```

msawyer-palantir: That would incorrectly write the bytes unmodified even if they use relocated classes

msawyer-palantir: To be clearer: relocation has to transform not just the relocated classes, but any classes that reference them as well.
We can only reuse the original class bytes if we walked the whole class file and didn't change anything, which this new callback tracks.

Goooler: You are right, I missed that.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowCopyAction.kt
```diff
@@ -185,14 +183,17 @@ public open class ShadowCopyAction(
      * Applies remapping to the given class with the specified relocation path. The remapped class is then written
      * to the zip file.
      */
-    private fun FileCopyDetails.remapClass() = file.inputStream().use { inputStream ->
+    private fun FileCopyDetails.remapClass() = file.readBytes().let { bytes ->
+      var hasModified = false
+      val remapper = RelocatorRemapper(relocators, onModified = { hasModified = true })
```
Goooler:
```suggestion
      val remapper = RelocatorRemapper(relocators) { hasModified = true }
```

Goooler (approved):

Issue #265, opened 2016-12-01
Add excluded dependencies to the shadow configuration automatically


Virtlink:
In my project I have to relocate a dependency because it clashes with a different version of the same dependency in the environment where my project is used. So I do:

```
dependencies {
    compile 'org.metaborg:org.metaborg.core:2.1.0-SNAPSHOT'
    compile 'commons-io:commons-io:2.4'
}

shadowJar {
    relocate 'org.apache.commons.io', 'shadow.org.apache.commons.io'
}
```

Now, because I relocated the dependency, I have to include in my shadowed jar all dependencies that themselves depend on the relocated dependency. But these dependencies transitively depend on other dependencies that I would like to exclude from the Jar. They should just be normal dependencies.

```
shadowJar {
    relocate 'org.apache.commons.io', 'shadow.org.apache.commons.io'

    dependencies {
        include(dependency('org.metaborg::'))
        include(dependency('commons-io:commons-io'))
   }
}
```

Doing this, I would expect the excluded (transitive) dependencies to be added to the `shadow` configuration automatically, since they are still dependencies my project or the shadowed jars. This appears not to be the case, and I would have to manually add all transitive dependencies to the `shadow` configuration. This is error prone, and as soon as one of my dependencies changes its own dependencies, I'd have to change my script.

Is there a way to add dependencies not included in the uber jar automatically to the `shadow` configuration?

johnrengelman:
I would add your dependencies to the `shadow` configuration instead and then only add the dependencies you want merged into `compile` and set `transitive = false` on them.

Virtlink:
So in this example it would become:

	dependencies {
		compile ('org.metaborg:org.metaborg.core:2.1.0-SNAPSHOT') {
			transitive=false
		}
		compile 'commons-io:commons-io:2.4'
		
		shadow 'org.metaborg:org.metaborg.core:2.1.0-SNAPSHOT'
	}

Wouldn't adding `org.metaborg.core` to the `shadow` configuration specify that my project needs `org.metaborg.core` as a normal dependency as well, while it's already included in the jar?

I believe what I'd like to write is something like:

	dependencies {
		compile ('org.metaborg:org.metaborg.core:2.1.0-SNAPSHOT') {
			transitive=false
		}
		compile 'commons-io:commons-io:2.4'
		
		shadow ('org.metaborg:org.metaborg.core:2.1.0-SNAPSHOT') {
			transitive=only
		}
	}

PR #1720, opened 2025-09-08
Move injecting `Class-Path` manifest attr logic from `doFirst` into `copy`

Goooler:
We need to move some manifest logic into `@TaskAction`, this is useful for #265.

---


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -496,16 +504,11 @@ public abstract class ShadowJar : Jar() {
...
-          if (!shadowFiles.isEmpty) {
-            val attrs = listOf(classPathAttr.get()) + shadowFiles.map { it.name }
-            task.manifest.attributes[classPathAttributeKey] = attrs.joinToString(" ").trim()
-          }
-        }
+        task.manifest = DefaultInheritManifest(
+          project,
+          @Suppress("EagerGradleConfiguration") // The ctor doesn't support Provider.
+          jarTask.get().manifest,
+        )
```
Goooler, line 511: This makes the regression test in #1782 fail.

PR #1773, opened 2025-09-20
Allow overriding `BUNDLING_ATTRIBUTE` in GMM

### LINKED ISSUE #1767: Non-relocating shadow jars should use `Bundling.EMBEDDED`
### Description

The `shadowRuntimeElements` is published with the publication using the `Bundling.SHADOWED` attribute. In Gradle's [documentation for Bundling](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.attributes/-bundling/index.html):

> - `external` indicates that transitive dependencies are themselves component jars
> - `embedded` indicates that transitive dependencies have been included inside the component jar, without modifying their packages
> - `shadowed` indicates that transitive dependencies have been included inside the component jar, under different packages to prevent conflicts

With that in mind, I think it would be a good idea to check, after evaluation, if the shadowJar task is relocating dependencies. If it is, use `SHADOWED`. And if it isn't, use `EMBEDDED`. This could be especially important since Shadow's default behavior is to not relocate packages.

Jonathing:
- Closes #1767.

Adds a new property to the `ShadowJar` task that dictates the bundling type to be used for the consumable configuration. I'm adding the bundling attribute using `AttributeContainer#attributeProvider` as lazily-added attribute aren't computed until they're needed, which is after project evaluation.

Once again looking for help with tests and code cleanup like last time. Kotlin is not my primary language :sweat_smile:

---

```kotlin
tasks.shadowJar {
  bundlingAttribute = 'embedded'
}
```

---

### DIFF
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.kt
@@ -26,7 +26,7 @@ public interface ShadowExtension {
    *
    * Per description of the attribute, you should set it to either [Bundling.SHADOWED] or [Bundling.EMBEDDED].
    *
-   * Defaults to `shadow`.
+   * Defaults to [Bundling.SHADOWED].
    */
   public val bundlingAttribute: Property<String>
 }
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
@@ -63,9 +63,7 @@ public abstract class ShadowJavaPlugin @Inject constructor(
         )
         attrs.attributeProvider(
           Bundling.BUNDLING_ATTRIBUTE,
-          shadow.bundlingAttribute.map { attr ->
-            objects.named(Bundling::class.java, attr)
-          },
+          shadow.bundlingAttribute.map { attr -> objects.named(Bundling::class.java, attr) },
         )
       }
       it.outgoing.artifact(tasks.shadowJar)
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
@@ -32,7 +32,6 @@ import org.apache.tools.zip.ZipOutputStream
 import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
-import org.gradle.api.attributes.Bundling
 import org.gradle.api.file.ArchiveOperations
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.CopySpec
--- a/docs/changes/README.md
+++ b/docs/changes/README.md
@@ -26,7 +26,7 @@
   You can override it for now by:
   ```kotlin
   shadow {
-    bundlingAttribute = org.gradle.api.attributes.Bundling.EMBEDDED
+    bundlingAttribute = Bundling.EMBEDDED
   }
   ```
 
--- a/docs/publishing/README.md
+++ b/docs/publishing/README.md
@@ -136,7 +136,7 @@ setting the `bundlingAttribute` property in the `shadow` extension:
     ```kotlin
     shadow {
       // Per description of the attribute, you should set it to either `Bundling.SHADOWED` or `Bundling.EMBEDDED`.
-      bundlingAttribute = org.gradle.api.attributes.Bundling.EMBEDDED
+      bundlingAttribute = Bundling.EMBEDDED
     }
     ```
 
@@ -145,7 +145,7 @@ setting the `bundlingAttribute` property in the `shadow` extension:
     ```groovy
     shadow {
       // Per description of the attribute, you should set it to either `Bundling.SHADOWED` or `Bundling.EMBEDDED`.
-      bundlingAttribute = org.gradle.api.attributes.Bundling.EMBEDDED
+      bundlingAttribute = Bundling.EMBEDDED
     }
     ```
 


Jonathing:
Oops, forgot to run tests before I clicked the submit button. Gimme a sec. Functional tests are working though.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowExtension.kt
```diff
@@ -19,4 +19,14 @@ public interface ShadowExtension {
...
+
+  /**
+   * If `true`, uses the [org.gradle.api.attributes.Bundling.EMBEDDED] attribute instead of the
+   * [org.gradle.api.attributes.Bundling.SHADOWED] attribute for the Gradle Module Metadata if the
+   * shadowed JAR does not relocate packages. This affects how consumers resolve the published artifact based on the
+   * requested [org.gradle.api.attributes.Bundling] attribute.
+   *
+   * Defaults to `false`.
+   */
+  public val useEmbeddedBundlingAttribute: Property<Boolean>
```
Goooler: I believe the new property could be 

```
val newProp: Property<String> // Defaults to `Bundling.SHADOWED`.
```

Users could easily override it.

Jonathing: That's fair, but the point is to lean into Gradle's idea of embedded being non-relocated packages. By having the plugin check that, it alleviates that burden off of the buildscript author. That's why I made it like this.

Jonathing: On another note, one thing I could do to make `ShadowJarPlugin.kt` less messy is move the work of checking if the task is relocating into `ShadowJar.kt`.

Goooler: Most users are not concerned about it. Advanced users need to fully control it.

Jonathing: Then an optional input property in the task could dictate a specific Bundling attribute to use, disabling the automatic check for whether to use embedded or shadowed.

Jonathing:
After thinking on your comments, I went ahead with a different approach, one that defaults to `Bundling.SHADOWED` and is completely dependent on the task itself rather than the `shadow` extension. Let me know if you got more comments or suggestions.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -244,6 +270,9 @@ public abstract class ShadowJar : Jar() {
    */
   override fun getDuplicatesStrategy(): DuplicatesStrategy = super.getDuplicatesStrategy()
 
+  @get:Inject
+  protected abstract val providers: ProviderFactory
```
Goooler: This is unnecessary.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -169,6 +172,29 @@ public abstract class ShadowJar : Jar() {
...
+  internal fun getFinalBundlingAttribute(): Provider<Bundling> {
+    return bundlingAttribute.orElse(
+      providers.zip<Set<Relocator>, Boolean, Boolean>(
+        relocators,
+        enableAutoRelocation,
+      ) { set, autoRelocating ->
+        !set.isEmpty() || autoRelocating
+      }.orElse(false).map { isRelocating -> if (isRelocating) Bundling.SHADOWED else Bundling.EMBEDDED },
+    ).map { objectFactory.named(Bundling::class.java, it) }
+  }
```
Goooler: We don't need this `getFinalBundlingAttribute`, it's a behavior change. If users want to override the default bundling property, just set `bundlingAttribute`.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -169,6 +170,17 @@ public abstract class ShadowJar : Jar() {
...
   public open val relocationPrefix: Property<String> = objectFactory.property(ShadowBasePlugin.SHADOW)
 
+  /**
+   * The [org.gradle.api.attributes.Bundling] attribute to use for the Gradle Module Metadata.
+   *
+   * If unset, Shadow will use [org.gradle.api.attributes.Bundling.SHADOWED] if this shadow JAR is relocating packages
+   * or [org.gradle.api.attributes.Bundling.EMBEDDED] if it is not.
+   *
+   * Defaults to `shadow`.
+   */
```
Goooler:
```suggestion
  /**
   * The [Bundling] attribute to use for the Gradle Module Metadata.
   *
   * Per description of the attribute, you should set it to either [Bundling.SHADOWED] or [Bundling.EMBEDDED].
   *
   * Defaults to `shadow`.
   */
```

Goooler:
> Once again looking for help with tests and code cleanup like last time. Kotlin is not my primary language 😅

You can add a new test ref the others in

https://github.com/GradleUp/shadow/blob/3efc02906a8f5afd9efc08947a2360ac85d89e44/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/PublishingTest.kt#L52

Also needs an extra check in https://github.com/GradleUp/shadow/blob/3efc02906a8f5afd9efc08947a2360ac85d89e44/src/test/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowPropertiesTest.kt#L90

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.kt
```diff
@@ -61,7 +62,10 @@ public abstract class ShadowJavaPlugin @Inject constructor(
           LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
           objects.named(LibraryElements::class.java, LibraryElements.JAR),
         )
-        attrs.attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling::class.java, Bundling.SHADOWED))
+
+        // Using AttributeContainer#attributeProvider means the value isn't queried until it is needed.
+        // Unless the attributes are consumed too early, this is an effective substitute for Project#afterEvaluate.
+        attrs.attributeProvider(Bundling.BUNDLING_ATTRIBUTE, tasks.shadowJar.map { task -> objects.named(Bundling::class.java, task.bundlingAttribute.get()) })
```
Goooler:
```suggestion
        attrs.attributeProvider(
          Bundling.BUNDLING_ATTRIBUTE,
          tasks.shadowJar.map { task ->
            objects.named(Bundling::class.java, task.bundlingAttribute.get())
          },
        )
```

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -169,6 +170,16 @@ public abstract class ShadowJar : Jar() {
...
 
+  /**
+   * The [Bundling] attribute to use for the Gradle Module Metadata.
+   *
+   * Per description of the attribute, you should set it to either [Bundling.SHADOWED] or [Bundling.EMBEDDED].
+   *
+   * Defaults to `shadow`.
+   */
+  @get:Internal
+  public open val bundlingAttribute: Property<String> = objectFactory.property(Bundling.SHADOWED)
```
Goooler: I believe this property could be moved into `ShadowExtension`, nothing special for `ShadowJar`.

Jonathing:
Thanks for finishing this up for me. I ended up having a busy start to my week.

PR #1830, opened 2025-10-30
Pin JDK release for `GroovyCompile`

Goooler:


Goooler:
Major version changed!

PR #1848, opened 2025-11-15
RFC: New transformers

snazy:
@Goooler I've created a couple new transformers. Would you mind taking a look which ones you think make sense for the project? I would then create individual PRs including tests for each of those.

* Deduplicating transformer - to include the same path for the same content only once, detects conflicts
* Merge property files transformer - to merge the properties of property files with the same path, detecting conflicts
* Merge license files transformer - to produce a license with with the project's license and the license files from the dependencies (but yea, no info about the actual Maven GAV)
* Enhance the Apache Notice transformer to make the output-path and source file patterns configurable
* Tool/convenience task to find dependencies that contain a matching pattern-filterable to debug conflicts.
* Helper class for the above task + transformers (hide boilerplate code)

PR #1851, opened 2025-11-17
Support overriding output path of `ApacheNoticeResourceTransformer`

Goooler:
Inspired by #1848.

---

### DIFF
--- a/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/TransformersTest.kt
+++ b/src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/TransformersTest.kt
@@ -251,6 +251,45 @@ class TransformersTest : BaseTransformerTest() {
     }
   }
 
+  @Test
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/ApacheNoticeResourceTransformer.kt
```diff
@@ -76,6 +76,14 @@ public open class ApacheNoticeResourceTransformer(
   @get:Input
   public open val charsetName: Property<String> = objectFactory.property(Charsets.UTF_8.name())
 
+  /**
+   * The output path of the `NOTICE` file.
+   *
+   * Defaults to `META-INF/NOTICE`.
+   */
+  @get:Input
+  public val outputPath: Property<String> = objectFactory.property(NOTICE_PATH)
```
snazy:
```suggestion
  public open val outputPath: Property<String> = objectFactory.property(NOTICE_PATH)
```
?

PR #1852, opened 2025-11-17
Directly call `setIncludes`

Goooler:


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/ServiceFileTransformer.kt
```diff
@@ -29,7 +29,6 @@ public open class ServiceFileTransformer @JvmOverloads constructor(
   @get:Internal
   internal val serviceEntries = mutableMapOf<String, MutableSet<String>>()
 
-  @Deprecated("Use `setIncludes` instead.", ReplaceWith("setIncludes()"))
   @get:Internal // No need to mark this as an input as `getIncludes` is already marked as `@Input`.
   public open var path: String = SERVICES_PATH
```
Goooler, line 33: I don't want to deprecate this setter for now, as it's handier than `setIncludes(listOf("$path/**"))`. But calling this will not reset `includes` fully. 

Should we document usages of `setIncludes` and `setExcludes`?

PR #1856, opened 2025-11-17
Add new merge strategy `Fail` to `PropertiesFileTransformer`

snazy:
Refs #1848.

---

- [x] [CHANGELOG](https://github.com/GradleUp/shadow/blob/main/docs/changes/README.md)'s "Unreleased" section has been updated, if applicable.

PR #1861, opened 2025-11-17
Make the output of `PropertiesFileTransformer` reproducible

snazy:
`PropertiesFileTransformer` leverages `java.util.Properties`, which relies on `java.util.Hashtable`. The serialized properties are not guaranteed to be reproducible.

This change changes the transformer to use a sorted map to generate reproducible output. The existing class `CleanProperties` is replaced with `ReproducibleProperties`.
Functionality around charset handling is retained, and extended with a functionality to generate unicode escapes (ASCII output).

Refs #1848.

---

- [X] [CHANGELOG](https://github.com/GradleUp/shadow/blob/main/docs/changes/README.md)'s "Unreleased" section has been updated, if applicable.


snazy:
Depends on #1856

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/ReproducibleProperties.kt
```diff
@@ -0,0 +1,60 @@
...
+  fun writeProperties(charset: Charset, os: OutputStream, escape: Boolean) {
+    val zipWriter = os.writer(charset)
+    props.forEach { (key, value) ->
+      zipWriter.write(convertString(key, isKey = true, escape))
+      zipWriter.write("=")
+      zipWriter.write(convertString(value, isKey = false, escape))
+      zipWriter.write("\n")
+    }
+    zipWriter.flush()
+  }
```
Goooler: This is complex than extending `CleanProperties` to be sorted, can we try something like https://stackoverflow.com/a/52127284/19772203? We just have to handle the sorting logic in `entrySet`.

snazy:
Okay, GH is broken :(

Goooler:
Let's wait to tomorrow 😁

snazy:
Oh, it's the "zulu" installer that barks.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/CleanProperties.kt
```diff
@@ -1,31 +1,40 @@
...
+    val bufferedReader = StringWriter().apply {
+      super.store(this, null)
+    }.toString().reader().buffered()
 
-    private val String?.couldBeCommentWithTimestamp: Boolean get() {
-      return this != null && startsWith("#") && length == lengthOfExpectedTimestamp
-    }
+    os.bufferedWriter(charset).apply {
+      var line: String? = null
+      while (bufferedReader.readLine().also { line = it } != null && line != null) {
```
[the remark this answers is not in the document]
Goooler: This is necessary for reducing `!!`.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/CleanProperties.kt
```diff
@@ -1,31 +1,40 @@
...
-      if (str.couldBeCommentWithTimestamp) return
-      super.write(str)
-    }
+  fun writeWithoutComments(charset: Charset, os: OutputStream) {
+    val bufferedReader = StringWriter().apply {
+      super.store(this, null)
+    }.toString().reader().buffered()
 
-    private val String?.couldBeCommentWithTimestamp: Boolean get() {
-      return this != null && startsWith("#") && length == lengthOfExpectedTimestamp
```
Goooler, line 28: The old logic is useful for removing date comments only. All comments are removed in this PR. See https://github.com/Goooler/shadow/pull/102#discussion_r1631130031.

I'm considering whether we should introduce a new property to opt for the feature in. The new flag will control:

1. Remove all comments or not.
2. Sort all properties or not.

Goooler (approved):

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/CleanProperties.kt
snazy: Not sure a flag's worth it.
The comment was always [empty](https://github.com/GradleUp/shadow/blob/57096c809eb9360dc2d46c0e1f6caf0926f4e1ef/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/transformers/PropertiesFileTransformer.kt#L240) (via [this](https://github.com/GradleUp/shadow/blob/57096c809eb9360dc2d46c0e1f6caf0926f4e1ef/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/Utils.kt#L57)).
Sorting vs not-sorting doesn't change the contents.

But we could (or should?) add a comment saying something like 
```
# Generated by GradleUp shadow plugin from one or more input properties files.
```
WDYT?

Goooler: > The comment was always empty

You are right!

> add a new comment...

Let's keep the current behavior; otherwise, we need to add comments for the property-related transformers.

Goooler:
I reverted the name change of `CleanProperties` to reduce the diff in this PR. You can send a new one for the renaming.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/CleanProperties.kt
Goooler: Updated a test for comments, see #1868.

Goooler:
Addressing the renaming to #1869.

PR #1871, opened 2025-11-20
Migrate to ktfmt

Goooler:


Goooler:
I don't really like this style.

PR #1885, opened 2025-12-10
Add Kdoc for applying resource transformers

### LINKED ISSUE #1884: Produce a warning when 'mergeServiceFiles()' is enabled with incompatible 'duplicatesStrategy'

According to [documentation](https://gradleup.com/shadow/configuration/merging/#handling-duplicates-strategy) `mergeServiceFiles()` does not work with `DuplicatesStrategy.EXCLUDE`, which is the default in `ShadowJar` task.

Please add a warning when `mergeServiceFiles()` is enabled and an incompatible `DuplicatesStrategy` is configured. This should reduce users' confusion on why service files were not merged and prevent long investigation times.

Maybe it should be a more generic solution for `ResourceTransform` - compatible `DuplicateStragegy` values are defined in the transform implementation, and the plugin produces warnings according to the list.

### Upstream changes

_No response_

_No response_

Goooler:
Closes #1884.

### DIFF
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
@@ -296,7 +296,8 @@ public abstract class ShadowJar : Jar() {
    * Merge Java services files with [rootPath].
    *
    * *Warning*: In most cases, this should be used with the correct [getDuplicatesStrategy] to ensure duplicate service
-   * files are handled properly.
+   * files are handled properly. See more details in the
+   * [Handling Duplicates Strategy](https://gradleup.com/shadow/configuration/merging/#handling-duplicates-strategy) section.
    *
    * @see [getDuplicatesStrategy]
    */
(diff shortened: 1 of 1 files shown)


Goooler:
CC @Tapchicoma

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -279,13 +279,23 @@ public abstract class ShadowJar : Jar() {
 
   /**
    * Merge Java services files with [rootPath].
+   *
+   * *Warning*: In most cases, this should be used with the correct [getDuplicatesStrategy] to ensure duplicate service
```
Tapchicoma, line 298: Could you also provide here a link to the documentation: https://gradleup.com/shadow/configuration/merging/#handling-duplicates-strategy ?

Goooler: ff512673f9ed5ea403ff5a1fa715e8fea6be1c0c

PR #1919, opened 2026-02-08
Gradle 9.4.0-rc-1

Goooler:
https://docs.gradle.org/9.4.0-rc-1/release-notes.html

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -204,7 +204,7 @@ gradlePlugin {
     }
   }
 
-  testSourceSets(sourceSets["functionalTest"], sourceSets["documentTest"])
+  testSourceSets(sourceSets["test"], sourceSets["functionalTest"], sourceSets["documentTest"])
 }
 
 // This part should be placed after testing.suites to ensure the test sourceSets are created.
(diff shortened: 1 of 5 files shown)


Goooler:
Blocker: https://github.com/gradle/gradle/issues/36652

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/GradleCompat.kt
```diff
@@ -79,6 +81,30 @@ internal fun AdhocComponentWithVariants.addVariantsFromConfigurationCompat(
   }
 }
 
+/** TODO: this could be removed after bumping the min Gradle requirement to 9.4 or above. */
+internal fun Project.moveGradleApiIntoCompileOnly() {
+  // gradleApi has been added into compileOnlyApi since Gradle 9.4-rc-1.
+  if (GradleVersion.current() < GradleVersion.version("9.4.0-rc-1")) return
```
big-guy: Isn't this check backwards?  You want to run this if the current version is older than 9.4.0-rc-1?  This only runs when the version is >= 9.4.0-rc-1.

Goooler: You are right! This is a typo 😅

PR #2003, opened 2026-04-10
Replace SetProperty<Configuration> with ConfigurableFileCollection on ShadowJar

tresat:
## Summary of Changes

This change replaces the `configurations` property on `ShadowJar` from `SetProperty<Configuration>` to `ConfigurableFileCollection`, and introduces three new public methods as the sole API for managing which Gradle configurations are shadowed:

- `addConfiguration(config: Configuration)` — add a configuration to shadow
- `setConfigurations(configs: Iterable<Configuration>)` — replace all configurations
- `clearConfigurations()` — remove all configurations

An internal `MutableList<Configuration>` (`_sourceConfigurations`) tracks the actual `Configuration` objects for dependency-level filtering, while the `ConfigurableFileCollection` serves as the configuration-cache-safe input property.

<details>
<summary>Files Modified</summary>

| File | Change |
|---|---|
| `ShadowJar.kt` | Replaced property type, added mutation methods, updated `toMinimize` and `includedDependencies` providers to reference internal list |
| `ShadowJavaPlugin.kt` | `task.configurations.convention(...)` replaced with `task.addConfiguration(runtimeConfiguration)` |
| `ShadowKmpPlugin.kt` | Same pattern — `task.addConfiguration(...)` replaces lazy convention |
| `ShadowPropertiesTest.kt` | Updated assertion from `configurations.get()` to `configurations.files` |
| `JavaPluginsTest.kt` | 4 Groovy DSL sites migrated to `setConfigurations([...])` |
| `RelocationTest.kt` | 1 Groovy DSL site migrated to `clearConfigurations()` |
| `PublishingTest.kt` | 1 Groovy DSL site migrated to `setConfigurations([...])` |

</details>

---

## The Problem: `Configuration` Cannot Survive a CC Round-Trip

`Configuration` cannot survive a configuration cache serialization round-trip. When the CC serializes a `Configuration` object, Gradle's codec produces a `FileCollection` (`ResolvingFileCollection`) on deserialization — not a `Configuration`. This causes a **task reconstruction failure at CC load time**, before the task ever executes.

The failure occurs when Gradle tries to restore a previously cached task. The CC deserializes the stored value (a `ResolvingFileCollection`) and attempts to inject it back into the `SetProperty<Configuration>`. The property performs a type check: "Is `ResolvingFileCollection` assignable to `Configuration`?" It is not — `ResolvingFileCollection` implements `FileCollection` but not `Configuration`. The build fails immediately:

```
Could not load the value of field 'configurations' of task ':shadowJar'
  > Cannot set the value of a property of type org.gradle.api.artifacts.Configuration
    using a provider of type
    org.gradle.api.internal.file.DefaultFileCollectionFactory$ResolvingFileCollection.
```

This is not a `.get()` call failure or a `ClassCastException`. The CC cannot even **reconstruct the task object**. The deserialization process fails during task restoration, which happens before any task action runs. Even if reconstruction somehow succeeded, calling `.get()` would return `FileCollection` objects where `Configuration` is expected — Shadow's `DependencyFilter` relies on `Configuration.resolvedConfiguration.firstLevelModuleDependencies` to filter by module coordinates, which would fail immediately.

Since [Gradle 8.0](https://docs.gradle.org/8.0/release-notes.html), the configuration cache always performs a store-then-load round-trip, even on the first build (a cache miss). Gradle configures the build, serializes the task graph, and then runs from the **deserialized** state — never from the live configuration objects. A type that cannot survive a round-trip will cause the build to fail on the very first invocation with the CC enabled.

---

## The Gradle PR: gradle/gradle#37466

[gradle/gradle#37466](https://github.com/gradle/gradle/pull/37466) addresses [gradle/gradle#19122](https://github.com/gradle/gradle/issues/19122) by moving the failure from a **confusing load-time type mismatch** to a **clear store-time rejection**.

The PR adds validation in the property codec encode paths (`PropertyCodec.encodeThis()`, `SetPropertyCodec.encodeThis()`, `ListPropertyCodec.encodeThis()`, `MapPropertyCodec.encodeThis()`) that checks whether the property's type argument is `Configuration` or `SourceDirectorySet`. If so, the build fails immediately at CC store time with:

```
Cannot serialize SetProperty<Configuration> in task :shadowJar of type ShadowJar.
The value type of this property (org.gradle.api.artifacts.Configuration)
is not supported with the configuration cache.

> Use a @InputFiles ConfigurableFileCollection instead.
```

### Key details

- **The validation fires for all annotations** — the PR tests `@Input`, `@Internal`, `@InputFiles`, and `@Classpath`. The `@Classpath` annotation does not bypass the check.
- **Unset properties are exempt** — if the `SetProperty<Configuration>` has never been set, its backing field is null, a null marker is written directly, and `SetPropertyCodec.encodeThis()` is never invoked.
- **The failure shifts from load to store** — this is strictly an improvement in error reporting. The underlying type-mismatch problem already existed; the PR just makes it impossible to write a poisoned cache entry.

| Aspect | Before PR #37466 | After PR #37466 |
|--------|-------------------|-----------------|
| CC Store | Succeeds silently | Fails with `UnsupportedPropertyValueException` |
| CC Load | Fails with confusing type mismatch | Never reached (poisoned entry not written) |
| Error message | "Cannot set the value of a property of type Configuration using a provider of type ResolvingFileCollection" | "Cannot serialize SetProperty\<Configuration\>... Use a @InputFiles ConfigurableFileCollection instead." |

---

## Why Shadow Works Today

Despite the round-trip problem described above, Gradle's own [Shadow plugin smoke test](https://github.com/gradle/gradle/blob/master/testing/smoke-test/src/smokeTest/groovy/org/gradle/smoketests/ShadowPluginSmokeTest.groovy) runs a full CC store-then-load cycle (build `shadowJar`, `clean`, build `shadowJar` again) and passes with the configuration cache enabled. The smoke test uses the plugin's default configuration — it does not set the `configurations` property explicitly.

The most likely explanation is that the `@Classpath` annotation causes the CC to handle this property through the **file-input serialization path** rather than through `SetPropertyCodec`. In this path, each `Configuration` (which extends `FileCollection`) is resolved to its constituent file paths at store time. On load, only the file-input tracking is restored — the CC does not attempt to reconstruct the `SetProperty<Configuration>` value by injecting a deserialized `FileCollection` back into it, so the type mismatch never occurs.

This is consistent with the PR's design: it adds rejection logic inside `SetPropertyCodec.encodeThis()`. If `@Classpath` properties bypass this codec entirely, the new rejection would not fire for Shadow's current property — but the PR's intent is to reject this pattern regardless, and future Gradle changes may close this bypass.

---

### The property replacement

```kotlin
// Before
@get:Classpath
public open val configurations: SetProperty<Configuration> = objectFactory.setProperty()

// After
@get:Classpath
public open val configurations: ConfigurableFileCollection = objectFactory.fileCollection()
```

`ConfigurableFileCollection` is the type Gradle's own error message recommends. The configuration cache serializes it as resolved file paths — no type mismatch, no round-trip problem. This resolves both the existing latent CC load failure and the new store-time rejection that gradle/gradle#37466 introduces.

### Preserving dependency filtering

The `DependencyFilter` contract requires actual `Configuration` objects to walk the dependency tree:

```kotlin
// DependencyFilter.kt
override fun resolve(configuration: Configuration): FileCollection {
  // Uses configuration.resolvedConfiguration.firstLevelModuleDependencies
  // to filter by group/name/version �� needs the real Configuration object
}
```

An internal `MutableList<Configuration>` (`_sourceConfigurations`) tracks the `Configuration` objects. The three public methods (`addConfiguration`, `setConfigurations`, `clearConfigurations`) serve as a chokepoint that keeps the file collection and the internal list in sync:

```kotlin
public open fun addConfiguration(config: Configuration) {
  _sourceConfigurations.add(config)    // for DependencyFilter
  configurations.from(config)           // for @Classpath / CC
}
```

### Why the internal list is safe on cache restore

`_sourceConfigurations` is annotated `@Transient` and is not a Gradle managed property — the configuration cache does not serialize it. After a cache restore, it's an empty list. This is safe because:

1. `includedDependencies` and `toMinimize` are `ConfigurableFileCollection` properties whose providers reference `_sourceConfigurations`
2. On the first run (cache store), these providers evaluate, call `DependencyFilter.resolve(_sourceConfigurations)`, and produce resolved file sets
3. The configuration cache serializes these `ConfigurableFileCollection`s with their resolved files
4. On cache restore, the resolved files are used directly — the providers are never re-evaluated, so `_sourceConfigurations` is never accessed

### DSL migration

The old `SetProperty` supported Groovy's property assignment syntax and Gradle's provider-aware `set(Provider<Iterable<T>>)`:

```groovy
// Old Groovy DSL
configurations = [project.configurations.runtimeClasspath]
configurations = project.configurations.named('testRuntimeClasspath').map { [it] }
```

The new API uses explicit method calls:

```groovy
// New Groovy DSL
setConfigurations([project.configurations.runtimeClasspath])
addConfiguration(project.configurations.testRuntimeClasspath)
clearConfigurations()
```

```kotlin
// New Kotlin DSL
setConfigurations(listOf(project.configurations.runtimeClasspath))
addConfiguration(project.configurations["testRuntimeClasspath"])
clearConfigurations()
```

This is a **breaking public API change** for users who configure `configurations` directly in their build scripts.


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -132,12 +135,34 @@ public abstract class ShadowJar : Jar() {
...
+    _sourceConfigurations.clear()
+    configurations.setFrom()
+    configs.forEach { addConfiguration(it) }
+  }
+
+  /** Remove all configurations. */
+  public open fun clearConfigurations() {
+    _sourceConfigurations.clear()
+    configurations.setFrom()
+  }
```
Goooler: I believe we don't have to add these new APIs after changing the type to `ConfigurableFileCollection`.

`configurations = listOf(someConfig)` works for Kotlin DSL, and it seems work for Groovy DSL as well.

tresat: If that works, it seems like a better way for you to move forward.

I put together this change because https://github.com/gradle/gradle/pull/37466 is something Gradle is currently strongly considering, and we wanted to test how difficult it would be to update Shadow.

Please do iterate on this and try to find something that works for your project.  Let us know if we can help.

Goooler: It works in ea3e9a52fbc17247c49d0c9a34b50b77ded74bdb.

docs/configuration/dependencies/README.md
```diff
@@ -9,15 +9,15 @@ merging can be configured using the [`configurations`][ShadowJar.configurations]
 
     ```kotlin
     tasks.shadowJar {
-      configurations = project.configurations.compileClasspath.map { listOf(it) }
+      configurations.setFrom(project.configurations.compileClasspath)
```
Goooler, line 12: I updated DSL samples like this, as there are extension overloads (`=` assignments) missing from Gradle side, errors like:

```
No applicable 'assign' function found for '=' overload
Argument type mismatch: actual type is 'NamedDomainObjectProvider<Configuration>', but 'FileCollection' was expected.
```

For the smooth migrations, I think we can add these extensions for Kotlin/Groovy DSL from Gradle side.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
tresat: I looked through it, and this looks like a better fix.

docs/configuration/dependencies/README.md
tresat: I don't think we'll be able to add these assignments for all `CFC`s in Gradle, if that's what you mean.  A `CFC` isn't generally equal to a list of `Configurations`.  This would have to be something specific to this plugin.

Goooler: Yes, but at least `configurations = project.configurations.compileClasspath` should work. It doesn't work at the moment, though. It reflects:

```
No applicable 'assign' function found for '=' overload
Argument type mismatch: actual type is 'NamedDomainObjectProvider<Configuration>', but 'FileCollection' was expected.
```

Goooler: Needs a overload like:

```kt
public fun ConfigurableFileCollection.assign(provider: Provider<*>) {
  setFrom(provider)
}
```

Refs https://github.com/gradle/gradle/blob/e1d3da2fd5259275fba81686a708f3c94d2bfc02/platforms/core-configuration/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/ConfigurableFileCollectionExtensions.kt#L61-L69.

PR #2077, opened 2026-07-01
Add support for R8-based minization

### LINKED ISSUE #885: Is it possible to specify the entry points for minimize configuration?
Use Case: The final jar contains multiple classes `A` and `B` from which the execution can start (these are the AWS lambda handlers).

Question: Is it possible to configure minimizing so that it would clean everything except `A`, `B` and their transitive dependencies?

### LINKED ISSUE #1176: Extend minimize() feature with support shrinkers/obfuscators like ProGuard/R8

Exitsitng [minimize() feature](https://gradleup.com/shadow/configuration/minimizing/) is very important for library authors to allow ship smaller as possible library.

But current implementation is very limited, doesn't provide enough control for unused dependencies control, doesn't support code optimization, obfuscation or code hardening (and it's understandable, all of those are out of scope for shadow plugin).

Because it obviously not realistic to include all potential features to shadow plugin, I propose to instead work on a pluggable API, which allows to write an additional plugin which will integrate shadow with existing obfuscation tools:
[ProGuard](https://github.com/Guardsquare/proguard)
[R8](https://developer.android.com/build/shrink-code)
[DexGuard](https://www.guardsquare.com/dexguard)
Many other security and optimization tools, too niche to support officially

It would be beneficial to have at least one official plugin for ProGuard or/and R8, so it would be easy for shadow users to find it and enable in their code. R8 is a good candidate, because it works for any JVM language (not only for Android), on active support, has Kotlin specific optimisations and probably even more globally adopted than ProGuard.

### Alternative solutions:
Do not integrate it on level of Shadow plugin but run them on top of jar.
The main issue is that it requires fully custom setup and anyway requires own gradle plugin, and it's not so trivial to do correctly. It doesn't improve shadow minimize feature, if any of external tools is used, it will be disabled anyway

### Related issues

https://github.com/GradleUp/shadow/issues/198
https://github.com/GradleUp/shadow/issues/565
https://github.com/GradleUp/shadow/issues/688

PR with R8 integration prototype:
https://github.com/GradleUp/shadow/pull/796 
Additionally, see the related discussion in previous PR: https://github.com/GradleUp/shadow/pull/566

### Upstream changes

_No response_

_No response_

ZacSweers:
This was attempted once or twice in the past on this project before the move, but attempted again here. This adds support for R8-based minimization, which is much more thorough and configurable than the default unused shrinker. This also allows for extra configuration to optimize, obfuscate, etc.

Full usage docs are in the README additions here. The default behavior is purely just tree-shaking. This also allows for the addition of other tools in the future too.

Disclosure: I did use AI to move through some of the boilerplate here, but manually de-fluffed and cleaned things up after too. Much of it was based on my use of R8 manually in another project here: https://github.com/ZacSweers/MoshiX/blob/main/moshi-ir/moshi-kotlin-tests/build.gradle.kts#L18

- Closes #885.
- Closes #1176.

---

### DIFF
--- a/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/R8Minimizer.kt
+++ b/src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/R8Minimizer.kt
@@ -2,7 +2,6 @@ package com.github.jengelman.gradle.plugins.shadow.internal
 
 import com.github.jengelman.gradle.plugins.shadow.relocation.Relocator
 import com.github.jengelman.gradle.plugins.shadow.relocation.relocateClass
-import com.github.jengelman.gradle.plugins.shadow.tasks.R8Spec
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
@@ -19,23 +18,24 @@ import org.gradle.process.ExecOperations
 /**
  * Runs R8 as a final-archive shrinker.
  *
- * Shadow first writes the complete jar with relocations, resource transformers, merged service
- * files, and duplicate handling applied. R8 then processes that exact artifact. The result is
- * normalized back through Shadow's archive settings because R8 does not know about this task's
- * reproducibility options.
+ * Shadow first writes the complete jar, including relocations, resource transformers, merged
+ * service files, and duplicate handling. R8 then processes that exact artifact.
  *
- * Generated rules intentionally use final jar contents: source-set classes are kept as roots,
- * dependencies excluded from minimization are kept, and entries under `META-INF/services` keep both
- * service interfaces and providers for downstream `ServiceLoader` use. User rule files and inline
- * rules are appended last. The default [R8Spec.args] are shrink-only: `--no-minification` also
- * causes Shadow to generate `-dontoptimize`; callers that replace the default args own the rest of
- * R8's behavior.
+ * R8 does not know about Shadow's reproducible archive settings, so its output is normalized before
+ * replacing the original jar.
+ *
+ * Generated rules are based on the final jar contents. Source-set classes are kept as roots,
+ * dependencies excluded from minimization are kept, and service descriptors keep providers for
+ * downstream `ServiceLoader` users. User rule files and inline rules are appended last.
+ *
+ * The default R8 configuration is shrink-only. Shadow passes `--no-minification` to disable name
+ * obfuscation and generates `-dontoptimize` unless optimization is enabled explicitly.
  */
 internal class R8Minimizer(
   private val execOperations: ExecOperations,
   private val logger: Logger,
   private val r8Classpath: FileCollection,
-  private val r8Spec: R8Spec,
+  private val r8Spec: DefaultR8Spec,
   private val sourceSetsClassesDirs: Iterable<File>,
   private val keptDependencyFiles: Iterable<File>,
   private val relocators: Iterable<Relocator>,
@@ -89,8 +89,8 @@ internal class R8Minimizer(
 
   private fun createRules(inputJar: File, r8Args: List<String>): List<String> {
     val rules = linkedSetOf<String>()
-    if (DefaultR8Spec.NO_MINIFICATION_ARG in r8Args) {
-      rules += "-dontoptimize"
+    if (shouldDisableOptimization(r8Args)) {
+      rules += DefaultR8Spec.DONT_OPTIMIZE_RULE
     }
     rules += sourceKeepRules(inputJar)
     rules += keptDependencyRules(inputJar)
@@ -104,6 +104,11 @@ internal class R8Minimizer(
       }
     rules += r8Spec.keepRules.get()
     return rules.toList()
+  }
+
+  private fun shouldDisableOptimization(r8Args: List<String>): Boolean {
+    if (r8Spec.optimizationEnabled.get()) return false
+    return r8Spec.obfuscationEnabled.get() || DefaultR8Spec.NO_MINIFICATION_ARG in r8Args
   }
 
   // Project classes are the public surface of the shadowed jar, even when nothing in the input jar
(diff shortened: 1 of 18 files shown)


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/MinimizeSpec.kt
```diff
@@ -0,0 +1,22 @@
...
+   *
+   * Defaults to [MinimizeTool.DEPENDENCY_ANALYZER].
+   */
+  @get:Input public val tool: Property<MinimizeTool>
+
+  /** Use R8 to minimize the shadowed JAR and configure its options. */
+  public fun r8(action: Action<in R8Spec>)
+
+  /** Use R8 to minimize the shadowed JAR and configure its options. */
+  public fun r8(action: Closure<*>)
```
Goooler: We don't have to expose `Closure`.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/R8Spec.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+
+/** Minimal R8 configuration for [ShadowJar.minimize]. */
+public interface R8Spec {
```
Goooler, line 13: Should we mark `DslMarker` for `r8` blocks?

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/R8Minimizer.kt
```diff
@@ -0,0 +1,337 @@
...
+      throw GradleException(
+        "R8 minimization requires a non-empty R8 classpath. Apply the Shadow plugin or configure the shadowR8 configuration."
+      )
+    }
+
+    val r8Dir = temporaryDir.resolve("r8").also { it.mkdirs() }
+    val rulesFile = r8Dir.resolve("rules.pro")
+    val r8Output = r8Dir.resolve("output.jar")
+    val normalizedOutput = r8Dir.resolve("normalized-output.jar")
+    val javaHome = System.getProperty("java.home")
```
Goooler: Should we support Java toolchian like `JavaExec`?

Goooler (approved):

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/MinimizeSpec.kt
Goooler: 14eeb12524d351c56a250bbcf7eb4fc751e0a18c

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/R8Minimizer.kt
ZacSweers: TIL! [f7ed650](https://github.com/GradleUp/shadow/pull/2077/commits/f7ed6506f21974d4c343f4960fc968b496782481)

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/R8Spec.kt
ZacSweers: [459da36](https://github.com/GradleUp/shadow/pull/2077/commits/459da3606b880b0a536a7592dd9cd580dd39e95a)

Goooler (approved):

Goooler:
Thanks a lot for this!

Supported auto-updates for R8 in #2080.

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/R8Spec.kt
```diff
@@ -0,0 +1,44 @@
...
+   */
+  @get:Input public val args: ListProperty<String>
+
+  /** Additional ProGuard/R8 keep rules. */
+  @get:Input public val keepRules: ListProperty<String>
+
+  /** Files containing additional ProGuard/R8 keep rules. */
+  @get:InputFiles
+  @get:PathSensitive(PathSensitivity.RELATIVE)
+  public val keepRuleFiles: ConfigurableFileCollection
```
Goooler, line 27: They should be renamed to `proguardRules` and `proguardRuleFiles`.

Goooler: #2120

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/MinimizeSpec.kt
```diff
@@ -0,0 +1,20 @@
...
+public interface MinimizeSpec : DependencyFilter {
+  /**
+   * The tool used to minimize the shadowed JAR.
+   *
+   * Defaults to [MinimizeTool.DEPENDENCY_ANALYZER].
+   */
+  @get:Input public val tool: Property<MinimizeTool>
+
+  /** Use R8 to minimize the shadowed JAR and configure its options. */
+  public fun r8(action: Action<in R8Spec>)
```
Goooler, line 19: After days of consideration, I would prefer there to be a separate `r8` block in `ShadowJar`, so that we can use `minimize` with `r8` separately.

Goooler: Would you mind taking a review for #2123?

PR #2089, opened 2026-07-06
Extract rules from dependency jars for R8 minification

ZacSweers:
This way libraries that embed their own proguard rules are respected

---

### DIFF
--- a/docs/changes/README.md
+++ b/docs/changes/README.md
@@ -5,7 +5,7 @@
 
 ### Added
 
-- Extract R8 rules from dependency JARs when using `minimize { r8 { ... } }`.
+- Extract R8 rules from dependency JARs when using `minimize { r8 { ... } }`. ([#2089](https://github.com/GradleUp/shadow/pull/2089))
 
 ## [9.5.1](https://github.com/GradleUp/shadow/releases/tag/9.5.1) - 2026-07-06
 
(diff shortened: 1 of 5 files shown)


docs/changes/README.md
```diff
@@ -3,6 +3,9 @@
 
 ## [Unreleased](https://github.com/GradleUp/shadow/compare/9.5.1...HEAD) - 2026-xx-xx
 
+### Added
+
+- Extract R8 rules from dependency JARs when using `minimize { r8 { ... } }`.
```
```suggestion
- Extract R8 rules from dependency JARs when using `minimize { r8 { ... } }`. ([#2089](https://github.com/GradleUp/shadow/pull/2089))
```

Goooler (approved):

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/R8Minimizer.kt
```diff
@@ -95,19 +101,50 @@ internal class R8Minimizer(
...
+    logger.info("Extracting R8 rules from {}.", inputJar)
+    execOperations.javaexec {
+      it.classpath = r8Classpath
+      it.mainClass.set(R8_RULES_MAIN_CLASS)
+      if (launcher != null) {
+        it.executable = launcher.executablePath.asFile.absolutePath
+      }
+      it.args(arguments)
+    }
+  }
```
Goooler, line 128: These embedded rules have already been honored by the R8 itself. See #2132.

PR #2094, opened 2026-07-08
Add note about output caching and inherited functions

bric3:
Clarify that inherited `CopySpec` functions disable output caching, indeed, most are actually invoking the `DefaultCopySpec.appendCopyAction` which disable output caching.

Changelog not updated as it's not a behavioral change.

---

### DIFF
--- a/docs/configuration/merging/README.md
+++ b/docs/configuration/merging/README.md
@@ -100,6 +100,8 @@ Alternatively, you can follow these steps:
     [`filesMatching`][Jar.filesMatching], [`filesNotMatching`][Jar.filesNotMatching], or [`eachFile`][Jar.eachFile] functions
     to set their `duplicatesStrategy` to `INCLUDE` or `WARN`.
 
+Note, however that functions inherited from [`CopySpec`][CopySpec], such as [`filesMatching`][Jar.filesMatching], [`filesNotMatching`][Jar.filesNotMatching],  [`eachFile`][Jar.eachFile], or others, disable the output caching.
+
 Optional steps:
 
 - Enable [`ShadowJar.failOnDuplicateEntries`][ShadowJar.failOnDuplicateEntries] to check duplicate entries in the final JAR.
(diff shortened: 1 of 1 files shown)


docs/configuration/merging/README.md
```diff
@@ -100,6 +100,8 @@ Alternatively, you can follow these steps:
     [`filesMatching`][Jar.filesMatching], [`filesNotMatching`][Jar.filesNotMatching], or [`eachFile`][Jar.eachFile] functions
     to set their `duplicatesStrategy` to `INCLUDE` or `WARN`.
 
+Note, however that functions inherited from [`CopySpec`][CopySpec], such as [`filesMatching`][Jar.filesMatching], [`filesNotMatching`][Jar.filesNotMatching],  [`eachFile`][Jar.eachFile], or others, disable the output caching.
```
Goooler, line 103: There is no footer link for `CopySpec`. And any source for 'disable the output caching'?

bric3: Fixed, thanks!

So I didn't found anything in user doc, but found the information inside the gradle source (9.6 at least)

https://github.com/gradle/gradle/blob/c5f8b7305a09bf706a0270593cdda9372e74ad48/platforms/core-configuration/file-operations/src/main/java/org/gradle/api/internal/file/copy/DefaultCopySpec.java#L510

The `DefaultCopySpec` methods like `fileMatching` end up calling eachFile which toggle a flag `hasCustomActions` that is consulted in `AbstractCopyTask` the parent to all archive tasks.

https://github.com/gradle/gradle/blob/c5f8b7305a09bf706a0270593cdda9372e74ad48/subprojects/core/src/main/java/org/gradle/api/tasks/AbstractCopyTask.java#L106-L109

bric3: I actually I walked history a bit and this is pretty old, it was introduced in Gradle 3.2

https://github.com/gradle/gradle/commit/510d032ec36b9302a6a026335f39042ced058c60

PR #2115, opened 2026-07-21
Preserve repeated lines in R8 rule files

ZacSweers:
Previously we were using a set, which would break multi-line rules since we read them line-by-line

### DIFF
--- a/docs/changes/README.md
+++ b/docs/changes/README.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
-- Preserve repeated lines in R8 rule files when using `minimize { r8 { ... } }`.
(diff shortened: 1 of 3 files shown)


docs/changes/README.md
```diff
@@ -6,6 +6,10 @@
 
 - Use `GradleException` for expected build failures. ([#2113](https://github.com/GradleUp/shadow/pull/2113))
 
+### Fixed
+
+- Preserve repeated lines in R8 rule files when using `minimize { r8 { ... } }`.
```
```suggestion
- Preserve repeated lines in R8 rule files when using `minimize { r8 { ... } }`. ([#2115](https://github.com/GradleUp/shadow/pull/2115))
```

Goooler (approved):

PR #2123, opened 2026-07-24
Decouple R8 from minimization with a deprecated compatibility DSL

Goooler:
Follow up https://github.com/GradleUp/shadow/pull/2077#discussion_r3642979849.

---


src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/tasks/ShadowJar.kt
```diff
@@ -711,18 +736,15 @@ public abstract class ShadowJar : Jar() {
...
-    if (!useR8) return
-    val keptDependencyFiles = includedDependencies.files - toMinimize.files
+  private fun runR8() {
+    if (!enableR8.get()) return
     R8Minimizer(
         execOperations = execOperations,
         logger = logger,
         r8Classpath = r8Classpath,
-        r8Spec = defaultMinimizeSpec.r8Spec,
+        r8Spec = defaultR8Spec as DefaultR8Spec,
```
Goooler:
```suggestion
        r8Spec = defaultR8Spec,
```

src/main/kotlin/com/github/jengelman/gradle/plugins/shadow/internal/R8Minimizer.kt
```diff
@@ -52,7 +50,7 @@ internal class R8Minimizer(
   fun minimize(inputJar: File, temporaryDir: File) {
     if (r8Classpath.isEmpty) {
       throw GradleException(
-        "R8 minimization requires a non-empty R8 classpath. Apply the Shadow plugin or configure the shadowR8 configuration."
+        "R8 post-processing requires a non-empty R8 classpath. Apply the Shadow plugin or configure the shadowR8 configuration."
       )
     }
```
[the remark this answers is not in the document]
Goooler: https://github.com/GradleUp/shadow/compare/1f73574c39444e29c7575a7d4ee3f71e44b51993..687ca72bf553d5b15719959c5d1154e5d71e6108

ZacSweers (commented):
I don't really see the value in this as it creates two different minimizing blocks rather than just one that supports different impls

src/functionalTest/kotlin/com/github/jengelman/gradle/plugins/shadow/MinimizeTest.kt
```diff
@@ -514,14 +566,16 @@ class MinimizeTest : BasePluginTest() {
...
     writeR8Repository()
     writeR8ClientAndServerModules(
       serverShadowBlock =
         """
         minimize {
           exclude(project(':client'))
-          r8 {}
+        }
+        r8 {
+          proguardRules.add("-keep class client.** { *; }")
```
Goooler, line 578: You're right. Running the minimizers independently requires users to duplicate exclusions as explicit R8 keep rules, while the existing `keptDependencyRules` handling already does this automatically.

Thanks for the feedback!

PR #2159, opened 2026-08-07
Exclude changelog file from documentTest task inputs

Goooler:


build.gradle.kts
```diff
@@ -228,11 +229,6 @@ buildConfig {
     buildConfigField("TEST_GRADLE_VERSION", testGradleVersion)
   }
   sourceSets.named("documentTest") {
-    val docsDir =
-      file("docs").also {
-        if (!it.exists() || !it.isDirectory)
-          error("Docs dir $it does not exist or is not a directory.")
-      }
```
Goooler, line 239: https://github.com/GradleUp/shadow/blob/9fec3d6c669c2192bdcab046b04fa76667cda47f/src/documentTest/kotlin/com/github/jengelman/gradle/plugins/shadow/DocCodeSnippetTest.kt#L17

PR #2275, opened 2026-09-03
Keep service implementations when service interface is used in minimize

Goooler:
Retain service provider implementations under `META-INF/services/` (and their transitive dependencies) when the corresponding service interface is kept during minimization.

Related

- https://issues.apache.org/jira/browse/MSHADE-313
- https://github.com/apache/maven-shade-plugin/pull/35

---

- [x] [CHANGELOG](https://github.com/GradleUp/shadow/blob/main/CHANGELOG.md)'s "Unreleased" section has been updated, if applicable.


Goooler:
Prefer using R8 instead. Documented in #2280.

Goooler:
This might be handled by https://github.com/tcurdt/jdependency/issues/385.

